PR #3, opened 2025-05-02
CPG flow implementation of the Dragen realginment pipeline

Alexander-Stuckey:
# Purpose

  - Refactor the Dragen realignment pipeline to use `cpg-flow` instead of `cpg-workflows`

## Proposed Changes

  - `cpg-workflow` -> `cpg-flow`
  - Refactor the `ManageDragenPipeline` stage into a CohortStage from a SequencingGroupStage
  - Implement error logging in GCP for samples that fail Dragen realignment

## Checklist

- [ ✅ ] Linting checks pass


MattWellie:
Screenshot 2025-05-02 at 12 13 33 PM
 🫠

.github/workflows/build_deploy_main.yaml
```diff
@@ -0,0 +1,91 @@
+name: Build and Deploy to main
```
MattWellie, line 1: you have 3 separate workflows all building docker files - you can condense this down to a single workflow, which runs under certain conditions, and conditionally pushes to different places. There's a few examples for this ([e.g.](https://github.com/populationgenomics/production-pipelines/blob/main/.github/workflows/docker.yaml#L58-L70))
- workflow runs when any of the conditions are met
- builds an image with a specific commit hash as a tag
- re-tags and pushes the image depending on which conditions were satisfied

Alexander-Stuckey: Yeah I know I need to clean that up. At least delete the `pull_request.yaml` as it adds nothing.

config/dragen_align_pa_defaults.toml
```diff
@@ -0,0 +1,64 @@
+# This is an example config file detailing all the optioons used for running the realignment pipeline
+[workflow]
+name = 'dragen_align_pa'
+output_prefix = "ica"
+sequencing_type = "genome"
+input_cohorts = [ "COH4965",]
+last_stages = ['DownloadDataFromIca']
```
MattWellie, line 7: The two last lines in this file need to be in the `[workflow]` block but they've been separated

```
# used to make sure we don't repeat previously completed stages
check_expected_outputs = true

# the method to register outputs, can be missing - will not generate metamist analysis entries
status_reporter = 'metamist'
```

.bumpversion.cfg
MattWellie, line 1: nice, should probably upstream this into the template repo. This bumpversion config stuff is migrated to the project TOML?

.github/workflows/build_deploy_main.yaml
Alexander-Stuckey: Going to leave the build on main separate for now, just so it's slightly easier for me to fix stuff when it breaks

src/dragen_align_pa/jobs/download_ica_pipeline_outputs.py
```diff
@@ -0,0 +1,59 @@
...
+from hailtop.batch.job import BashJob
+from loguru import logger
+
+
+def _initalise_bulk_download_job(sequencing_group: SequencingGroup) -> BashJob:
+    bulk_download_job = get_batch().new_bash_job(
+        name='DownloadDataFromIca',
+        attributes=(sequencing_group.get_job_attrs() or {}) | {'tool': 'ICA'},  # type: ignore  # noqa: PGH003
+    )
+    bulk_download_job.image(image=config_retrieve(['images', 'ica']))
```
MattWellie, line 16: I can see this causing some problems at some point. The current expectation is that the driver & child jobs use the same image (I think?), which is guaranteed by `config.workflow.driver_image`. 

By instead using `config_retrieve(['images', 'ica'])` you create issues unless the config is updated with every pipeline change (and also means you would need to update the config with the path to a test image on a dev run etc.).

This works, but so long as the image is the same for these tasks `config.workflow.driver_image` is zero effort for the same result.

src/dragen_align_pa/stages.py
```diff
@@ -0,0 +1,345 @@
...
+set +x
+gcloud secrets versions access latest --secret=illumina_cpg_workbench_api --project=cpg-common | jq -r .apiKey > key
+gcloud secrets versions access latest --secret=illumina_cpg_workbench_api --project=cpg-common | jq -r .projectID > projectID
+echo "x-api-key: $(cat key)" >> $HOME/.icav2/config.yaml
+icav2 projects enter $(cat projectID)
+set -x
+"""  # noqa: E501
+
+
+def calculate_needed_storage(
```
MattWellie, line 46: this method was moved to utils, isn't used here

src/dragen_align_pa/stages.py
```diff
@@ -0,0 +1,345 @@
...
+        return sg_bucket / GCP_FOLDER_FOR_ICA_PREP / f'{sequencing_group.name}_output_fid.json'
+
+    def queue_jobs(self, sequencing_group: SequencingGroup, inputs: StageInput) -> StageOutput | None:  # noqa: ARG002
+        bucket_name: str = get_path_components_from_gcp_path(path=str(object=sequencing_group.cram))['bucket']
+        outputs: cpg_utils.Path = self.expected_outputs(sequencing_group=sequencing_group)
+
+        ica_prep_job: PythonJob = prepare_ica_for_analysis.run_ica_prep_job(
+            sequencing_group=sequencing_group,
+            output=str(outputs),
+            ica_analysis_output_folder=config_retrieve(['ica', 'data_prep', 'output_folder']),
```
MattWellie, line 73: nit: slight inconsistency here, you're getting this value in the stages.py here, but for two other stages you're pulling this value from config in the job instead

src/dragen_align_pa/stages.py
```diff
@@ -0,0 +1,345 @@
...
+
+        return self.make_outputs(
+            target=sequencing_group,
+            data=output,
+            jobs=upload_job,
+        )
+
+
+@stage(
+    required_stages=[PrepareIcaForDragenAnalysis, UploadDataToIca],  # type: ignore  # noqa: PGH003
```
MattWellie, line 109:
```suggestion
    required_stages=[PrepareIcaForDragenAnalysis, UploadDataToIca],  # noqa: PGH003
```

I think `type: ignore` is a universal ignore, `noqa: CODE` is a specific ignore. Not sure stacking them is necessary

src/dragen_align_pa/stages.py
```diff
@@ -0,0 +1,345 @@
...
+                for sequencing_group in cohort.get_sequencing_groups()
+            }
+            | {
+                f'{sequencing_group.name}_pipeline_id': sg_bucket
+                / GCP_FOLDER_FOR_RUNNING_PIPELINE
+                / f'{sequencing_group.name}_pipeline_id.txt'
+                for sequencing_group in cohort.get_sequencing_groups()
+            }
+            | {f'{cohort.name}_errors': sg_bucket / GCP_FOLDER_FOR_RUNNING_PIPELINE / f'{cohort.name}_errors.log'}
+        )
```
MattWellie, line 143:
```suggestion
        prefix: cpg_utils.Path = sg_bucket / GCP_FOLDER_FOR_RUNNING_PIPELINE
        results: dict[str, cpg_utils.Path] = {f'{cohort.name}_errors': prefix / f'{cohort.name}_errors.log'}
        for sequencing_group in cohort.get_sequencing_groups():
            results |= {
                f'{sequencing_group.name}_success': prefix / f'{sequencing_group.name}_pipeline_success.json',
                f'{sequencing_group.name}_pipeline_id': prefix / f'{sequencing_group.name}_pipeline_id.txt',
            }
        return results
```

This syntax is an affront to god and man

src/dragen_align_pa/stages.py
```diff
@@ -0,0 +1,345 @@
...
+
+        ica_download_job: BashJob = download_specific_files_from_ica.download_data_from_ica(
+            job_name='DownloadGvcfFromIca',
+            sequencing_group=sequencing_group,
+            filetype='gvcf',
+            bucket=get_path_components_from_gcp_path(path=str(object=sequencing_group.cram))['bucket'],
+            ica_cli_setup=ICA_CLI_SETUP,
+            gcp_folder_for_ica_download=GCP_FOLDER_FOR_ICA_DOWNLOAD,
+            pipeline_id_path=str(pipeline_id_path),
```
MattWellie, line 292: pipeline_id_path is type annotated as a Path but cast as a String twice

src/dragen_align_pa/stages.py
```diff
@@ -0,0 +1,345 @@
...
+        sequencing_group: SequencingGroup,
+    ) -> cpg_utils.Path:
+        bucket_name: cpg_utils.Path = sequencing_group.dataset.prefix()
+        return bucket_name / GCP_FOLDER_FOR_ICA_DOWNLOAD / 'dragen_metrics' / f'{sequencing_group.name}'
+
+    def queue_jobs(self, sequencing_group: SequencingGroup, inputs: StageInput) -> StageOutput | None:
+        outputs: cpg_utils.Path = self.expected_outputs(sequencing_group=sequencing_group)
+
+        # Inputs from previous stage
+        pipeline_id_path: cpg_utils.Path = str(
```
MattWellie, line 324: same here (wrong type annotation/multiple str casts)

MattWellie (changes requested):
nice

.bumpversion.cfg
Alexander-Stuckey: Yes it has although don't know if that means you need to use `bump-my-version` over the older methods (which I think are now unsupported anyway)

config/dragen_align_pa_defaults.toml
Alexander-Stuckey: Ah that'd be why it kept rerunning finished stages.

src/dragen_align_pa/stages.py
Alexander-Stuckey: Yes, I meant to fix that

src/dragen_align_pa/stages.py
Alexander-Stuckey: Think I'll clean this up later tbh, find the correct ignore code

src/dragen_align_pa/stages.py
Alexander-Stuckey: You work with what you've got I guess

src/dragen_align_pa/jobs/download_ica_pipeline_outputs.py
Alexander-Stuckey: dev images get overwritten, and `bump-my-version` does change the version in the config. Of course, that relies on it being run on a PR.

I can look into cleaning this up a bit later, since it works for now.

Alexander-Stuckey: Do you mean that `config.workflow.driver_image` is set to the image that is submitted to `analysis-runner`? Because that is better yeah

Alexander-Stuckey: Ok yes this is a much better implementation

Alexander-Stuckey:
Any further comments @MattWellie @michael-harper @katiedelange , or does it all look good for an approval?

MattWellie:
I'll have another quick skim, looked fine to me but thought the final stamp should come from your team

Dockerfile
```diff
@@ -1 +1,34 @@
...
+
+
+WORKDIR /dragen_align_pa
+
+# Add in the additional requirements that are most likely to change.
+COPY LICENSE pyproject.toml README.md .
+COPY src src/
+
+RUN pip install git+https://github.com/Illumina/ica-sdk-python.git \
+    && pip install loguru \
```
MattWellie, line 32: Feels like loguru should be in the project's dependencies instead of installed separately.

Currently this project lists no dependencies outside of cpg-flow which is fine, but if you wait for [this](https://github.com/populationgenomics/cpg-flow/actions/runs/14830832525) to be resolved the latest cpg-flow version (0.2.2?) will also have loguru as a dependency so you can drop this line.

Alexander-Stuckey: That's a good shout

pyproject.toml
```diff
@@ -1,15 +1,17 @@
...
+# requires = ["setuptools"]
+# build-backend = "setuptools.build_meta"
+requires = ["hatchling >= 1.26"]
+build-backend = "hatchling.build"
 
 [project]
-# TODO(you): update this to reflect the project's name & description
-name='workflow_name'
-description='Template workflow'
+# # TODO(you): update this to reflect the project's name & description
```
MattWellie, line 8:
```suggestion
```

src/dragen_align_pa/stages.py
```diff
@@ -0,0 +1,327 @@
...
+        prefix: cpg_utils.Path = sg_bucket / GCP_FOLDER_FOR_RUNNING_PIPELINE
+        results: dict[str, cpg_utils.Path] = {f'{cohort.name}_errors': prefix / f'{cohort.name}_errors.log'}
+        for sequencing_group in cohort.get_sequencing_groups():
+            results |= {
+                f'{sequencing_group.name}_success': prefix / f'{sequencing_group.name}_pipeline_success.json',
+                f'{sequencing_group.name}_pipeline_id': prefix / f'{sequencing_group.name}_pipeline_id.txt',
+            }
+        return results
+
+    def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:
```
MattWellie, line 133:
```suggestion
    def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput:
```

`StageOutput | None` is the exact method template in cpg-flow, but this will never return None

MattWellie: happens a couple of times, but writing this once

src/dragen_align_pa/stages.py
```diff
@@ -0,0 +1,327 @@
...
+        """  # noqa: E501
+        outputs: dict[str, cpg_utils.Path] = self.expected_outputs(sequencing_group=sequencing_group)
+
+        # Inputs from previous stage
+        pipeline_id_path: cpg_utils.Path = list(  # noqa: RUF015
+            inputs.as_path_by_target(
+                stage=ManageDragenPipeline,  # type: ignore[ReportUnknownVariableType]
+                key=f'{sequencing_group.name}_pipeline_id',
+            ).values()
+        )[0]
```
MattWellie, line 267: this is funky as heck. I mean... I guess it works, but it only works at exactly one Cohort per run. Is that documented anywhere?

MattWellie (approved):
Couple of pointless comments, generally looks good

michael-harper (approved):
This looks good to me. Thanks @MattWellie for the thoroughness!

src/dragen_align_pa/stages.py
Alexander-Stuckey: Yeah this is crimes against code. We only ever run a single cohort through the pipeline at any time, so it will be fine. I'll make sure that the readme reflects this though

src/dragen_align_pa/stages.py
Alexander-Stuckey: Fixed, will clean up the MLR stages when I write them.

katiedelange (approved):
I have not reviewed the bulk of the code here, as I assume that review has been well covered by Matt and Harper (and I reviewed the previous pipeline before this cpg-flow migration in detail).

Instead, I have only specifically checked the README and example config file, both of which look good. Nice job Alex!

PR #6, opened 2025-05-08
write error logs if both failed and cancelled pipelines

Alexander-Stuckey:
# Purpose

  - Fix issue with the error log not getting written if there are both failed and cancelled pipeline runs in the same management job.
  
## Proposed Changes

  - If there are pipeline failures as well as cancelled pipeline runs, make sure to write the error log to GCP first, before raising the exception about cancelled pipelines.


MattWellie (approved):
LGTMatt

PR #9, opened 2025-06-12
Workaround for bioheart

Alexander-Stuckey:
# Purpose

Bioheart was run through alignment and genotyping before the `ar_guid` was introduced. This patch allows the pipeline to work on Bioheart data, while still retaining the proper `ar_guid` structure for all future runs.

## Proposed Changes
  - Patch to run on older Bioheart runs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM australia-southeast1-docker.pkg.dev/cpg-common/images/cpg_hail_gcloud:0.2.134.cpg1
 
 # Dragen align pa pipeline version.
-ENV VERSION=2.0.0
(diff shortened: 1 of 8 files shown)


Alexander-Stuckey:
> this gives me a headache, but will probably work (?)
> 
> Main question is - is this true of every BioHeart sample now and forever? Or will we potentially have a mix of ARG/non-ARG at some point

It's only for the samples that have already been run. Once those have been processed through the MLR step, and all the data downloaded and deleted from ICA, I plan to revert these changes.

MattWellie (approved):
In AS we trust 🫡

PR #12, opened 2025-06-26
Address review comments

Alexander-Stuckey:
# Purpose

  - Fix the cancelling logic in both Dragen and MLR stages
  - Always check all sequencing groups in each iteration through to look for cancellations and large numbers of failures
  - Add in logic to retry the Dragen alig and genotype once in the case of transient Dragen errors

## Proposed Changes

  - Fix cancleeing logic
  - Add retry logic

### DIFF
--- a/src/dragen_align_pa/jobs/manage_dragen_mlr.py
+++ b/src/dragen_align_pa/jobs/manage_dragen_mlr.py
@@ -159,8 +159,24 @@ def _run(  # noqa: PLR0915
             ):
                 continue
 
+            # Get an MLR analysis ID for each sequencing group, if it exists.
+            mlr_analysis_id: str = ''
+            if mlr_pipeline_id_file_exists := mlr_pipeline_id_file.exists():
+                with mlr_pipeline_id_file.open('r') as mlr_pid_fhandle:
+                    mlr_analysis_id = json.load(mlr_pid_fhandle)['pipeline_id']
+
+            # Handle cancelling a pipeline first
+            # Cancel a running job in ICA
+            if config_retrieve(key=['ica', 'management', 'cancel_cohort_run'], default=False) and mlr_analysis_id:
+                logger.info(f'Cancelling pipeline run: {mlr_analysis_id} for sequencing group {sg_name}')
+                cancel_ica_pipeline_run.run(ica_pipeline_id=mlr_analysis_id, api_root=api_root, is_mlr=True)
+                delete_pipeline_id_file(pipeline_id_file=str(mlr_pipeline_id_file))
+            else:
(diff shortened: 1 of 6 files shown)


src/dragen_align_pa/jobs/manage_dragen_mlr.py
```diff
@@ -219,28 +226,24 @@ def _run(  # noqa: PLR0915
...
+                        lines: list[str] = tmp_log_handle.readlines()
+                        with outputs[f'{cohort.name}_errors'].open('a') as gcp_error_log_file:
+                            gcp_error_log_file.write('\n'.join(lines))
+                raise Exception(f'The following MLR pipelines have been cancelled: {" ".join(cancelled_pipelines)}')
+
+            # If more than 5% of pipelines are failing, exit now so we can investigate
+            if failed_pipelines and float(len(failed_pipelines)) / float(len(cohort.get_sequencing_groups())) > 0.05:  # noqa: PLR2004
+                raise Exception(
+                    f'More than 5% of MLR pipelines have failed. Failing pipelines: {" ".join(failed_pipelines)}'
+                )
```
katiedelange, line 246: I don't think the new indent was intended here? The `failed_pipelines` check can reasonably happen inside the for loop as you are just raising an exception and exiting whenever that threshold of 5% is reached, not writing out the failures, but the cancelled & failed pipelines check where you write out the failed pipelines to a log file needs to be after processing all sequencing groups.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
```diff
@@ -139,28 +148,38 @@ def _run(  # noqa: PLR0915
...
+                        with outputs[f'{cohort.name}_errors'].open('a') as gcp_error_log_file:
+                            gcp_error_log_file.write('\n'.join(lines))
+                raise Exception(f'The following pipelines have been cancelled: {" ".join(cancelled_pipelines)}')
+
+            # If more than 5% of pipelines are failing, exit now so we can investigate
+            if failed_pipelines and float(len(failed_pipelines)) / float(len(cohort.get_sequencing_groups())) > 0.05:  # noqa: PLR2004
+                raise Exception(
+                    f'More than 5% of pipelines have failed. Failing pipelines: {" ".join(failed_pipelines)}'
+                )
+            # Catch just in case everything is finished on the first pass through the loop
```
katiedelange, line 182: As per above regarding indents.

src/dragen_align_pa/jobs/manage_dragen_mlr.py
Alexander-Stuckey: Yes, my brain is having a lot of issues today

PR #21, opened 2025-09-14
Non preempt for data copy

Alexander-Stuckey:
# Purpose

  - Swap to using non-preemptible machines for data copy operations, so we only ever pay data egress charges once per file.


michael-harper (approved):
LGTM

PR #23, opened 2025-10-01
install procps so multiqc works

Alexander-Stuckey:
# Purpose

  - MultiQC doesn't work. Make it work

## Proposed Changes

  - Install `procps`


michael-harper (approved):
LGTM

PR #24, opened 2025-10-30
Update prepare ica

Alexander-Stuckey:
**Title: Feat: Refactor ICA state management, fix FASTQ inputs, and enhance cleanup**

### Purpose

This PR introduces a major refactor to how the pipeline manages external ICA jobs, moving from bespoke logic in each stage to a generic, reusable state machine. This change significantly improves the pipeline's robustness, resumability, and maintainability.

This update also fixes a critical bug that caused the FASTQ pipeline to fail. The input contract for QC BED files was incorrect, leading to the `Could not open BED file` error.

Finally, this PR enhances the `DeleteDataInIca` stage to fully clean up source data, not just analysis outputs, to better manage ICA storage costs.

### Proposed Changes

* **Refactored ICA State Management:**
    * Introduced a generic, reusable state-management loop in `src/dragen_align_pa/jobs/ica_pipeline_manager.py`.
    * This new manager centralizes all logic for submitting, monitoring, resuming, and cancelling external ICA jobs.
    * Refactored the `ManageDragenPipeline`, `ManageDragenMlr`, and `FastqIntakeQc` (via `manage_md5_pipeline.py`) stages to use this new, robust manager.

* **Bug Fix: Correct FASTQ Input Contract:**
    * Fixed the bug in `src/dragen_align_pa/jobs/run_align_genotype_with_dragen.py` that caused the `exit code 134` (SIGABRT) error.
    * The code now correctly passes QC BED files using the `qc_coverage_region_beds` (plural, multi-file) parameter for FASTQ runs, as defined in its XML contract. This is distinct from the CRAM pipeline's `qc_coverage_region_1` / `_2` (singular, single-file) parameters.

* **Enhanced Data Cleanup:**
    * The `DeleteDataInIca` stage is now a `CohortStage`.
    * The stage logic is updated to delete all original source data (uploaded CRAMs or FASTQs) in addition to the generated analysis folders, providing a complete cleanup of the ICA project.

* **Documentation Overhaul:**
    * Added a new `README.md` with a workflow DAG, aimed at junior developers.
    * Added a technical `README_developer.md` and a high-level `README_lead.md` for different audiences.

* **Dependency & Utility Updates:**
    * Upgraded the vendored `popgen-cli` from v2.0.0 to v2.1.0.
    * Refactored `somalier_extract.py` and `run_multiqc.py` to use more robust `gcloud storage` commands for file operations.
    * Updated `ica_utils.py` with several new helper functions to support the refactored download and state management logic.


src/dragen_align_pa/stages.py
```diff
@@ -11,100 +12,225 @@
...
+        return intake_qc_results
+
+    def queue_jobs(self, cohort: Cohort, inputs: StageInput) -> StageOutput | None:  # noqa: ARG002
+        if READS_TYPE == 'fastq':
+            outputs: dict[str, cpg_utils.Path] = self.expected_outputs(cohort=cohort)
+            md5job: PythonJob = manage_md5_pipeline.run_md5_management_job(cohort=cohort, outputs=outputs)
+
+            return self.make_outputs(target=cohort, data=outputs, jobs=md5job)  # pyright: ignore[reportArgumentType]
+
+        return None
```
michael-harper: I think we should raise an error here that tells the user they forgot to set the config `['workflow', 'reads_type']` as fastq instead of the workflow likely just saying 'no jobs to submit'. If raising an error it will get printed to the batch screen and can be quickly fixed.

src/dragen_align_pa/jobs/prepare_ica_for_analysis.py
```diff
@@ -1,85 +1,75 @@
...
 
 
 def _run(
-    ica_analysis_output_folder: str,
-    api_root: str,
-    sg_name: str,
-    bucket_name: str,
-) -> dict[str, str]:
+    cohort: Cohort,
+    output: dict[str, cpg_utils.Path],
```
michael-harper:
```suggestion
    outputs: dict[str, cpg_utils.Path],
```

src/dragen_align_pa/stages.py
```diff
@@ -11,100 +12,225 @@
...
+
+@stage(required_stages=[FastqIntakeQc])
+class DownloadMd5Results(CohortStage):
+    """
+    Downloads the 'all_md5.txt' result file from a successful
+    MD5 Checksum pipeline run.
+    """
+
+    def expected_outputs(self, cohort: Cohort) -> dict[str, cpg_utils.Path]:  # pyright: ignore[reportIncompatibleMethodOverride]
+        return {'ica_md5sum_file': get_prep_path(filename=f'{cohort.name}_ica_md5sum.md5sum')}
```
michael-harper: Nitpick but doesn't necessarily need to be a dict if just returning one expected output

src/dragen_align_pa/jobs/make_fastq_file_list.py
```diff
@@ -0,0 +1,118 @@
...
+    job.call(_run, outputs=outputs, cohort=cohort)
+
+    return job
+
+
+def _run(
+    outputs: dict[str, cpg_utils.Path],
+    cohort: Cohort,
+) -> None:
+    # Somtimes the contents of sequiencing_group.assays.meta['reads'] is a nested list
```
michael-harper:
```suggestion
    # Somtimes the contents of sequencing_group.assays.meta['reads'] is a nested list
```

src/dragen_align_pa/jobs/run_align_genotype_with_dragen.py
```diff
@@ -112,37 +279,24 @@ def run(
         output_path (str): The path to write the pipeline ID to
     """  # noqa: E501
```
michael-harper: Update `Args` above to reflect the arguments actually being passed.

src/dragen_align_pa/stages.py
```diff
@@ -336,25 +554,31 @@ def queue_jobs(self, sequencing_group: SequencingGroup, inputs: StageInput) -> S
...
             target=sequencing_group,
-            data=outputs,
+            data=outputs,  # pyright: ignore[reportArgumentType]
             jobs=ica_download_job,
         )
 
 
 @stage(
-    analysis_type='ica_data_download',
     required_stages=[
```
michael-harper: I think we should discuss whether to run MultiQC as part of this pipeline or the pre-combiner sample qc pipeline. We download the `dragen_metrics` folder from ICA anyway so we can delete them from there as necessary once alignment and genotyping is done.

This doesn't feel like a QC pipeline.

michael-harper (changes requested):
Very clean! A few comments here and there, nothing that I think requires too much of a change and are probably due to a misunderstanding of how the data looks. 

Further discussion needed for whether or not this is the best place for MultiQC as well.

Looks good!

src/dragen_align_pa/stages.py
Alexander-Stuckey: I have added this check into `constants.py`

src/dragen_align_pa/stages.py
```diff
@@ -11,100 +12,225 @@
...
         return self.make_outputs(
-            target=sequencing_group,
-            data=outputs,
+            target=cohort,
+            data=outputs,  # pyright: ignore[reportArgumentType]
             jobs=ica_prep_job,
         )
 
 
+@stage(required_stages=[PrepareIcaForDragenAnalysis])
```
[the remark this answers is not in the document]
Alexander-Stuckey: I'm 99% sure it reuses the output folder for the cohort as the base folder to write the md5 results into. It's possible I could disconnect the stages, but this it's possible that both stages try to create the ICA path at the same time and throw an error in ICA. This way is a little safer, and doesn't really cost any time.

src/dragen_align_pa/jobs/validate_md5_sums.py
```diff
@@ -0,0 +1,136 @@
...
+    # All files validated successfully, write the success file
+    logger.info('All MD5 checksums validated successfully.')
+    try:
+        with success_output_path.open('w') as success_fh:
+            success_fh.write('SUCCESS')
+        logger.info(f'Success file written to {success_output_path}')
+    except Exception as success_e:
+        logger.error(f'Failed to write MD5 success file to {success_output_path}: {success_e}')
+        # Raise an error here as well, because the stage expects this file
+        raise
```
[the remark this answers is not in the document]
Alexander-Stuckey: The ICA pipeline has 3 retries, so in theory all md5sums will be calculated. If one doesn't, the ICA pipeline should throw an error. There's a lot of error checking here in case something does go wrong though.

This particular exception is just in case everything goes well but then we fail to write the results out to gcp.

src/dragen_align_pa/jobs/ica_pipeline_manager.py
```diff
@@ -0,0 +1,216 @@
...
+                                 called, submits the job and returns the
+                                 pipeline_id (str).
+        allow_retry: Whether to retry a failed pipeline once.
+        sleep_time_seconds: Time to sleep between polling loops.
+    """
+    if targets_to_process:
+        # Explicitly get the first target after checking the sequence is not empty
+        first_target = targets_to_process[0]
+        # Both Cohort and SequencingGroup have a .name attribute
+        cohort_name = first_target.name
```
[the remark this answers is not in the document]
Alexander-Stuckey: Yeah here `cohort_name` doesn't specifically reference a Metamist cohort object (confusing :( ) . It's just representing a collection of targets to process. I have changed the variable name.

src/dragen_align_pa/jobs/manage_md5_pipeline.py
```diff
@@ -0,0 +1,207 @@
...
+    This function runs inside the PythonJob.
+    It performs the pre-submission setup (getting FASTQ IDs, creating folders)
+    and then calls the generic pipeline manager.
+
+    """
+
+    cohort_name: str = cohort.name
+    manifest_file_path: cpg_utils.Path = config_retrieve(
+        ['workflow', 'manifest_gcp_path'],
+    )
```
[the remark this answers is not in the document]
Alexander-Stuckey: Updated in base readme. I don't think that these will change very often, if at all. They are created as a part of data ingestion.

michael-harper (approved):
LGTM!

PR #26, opened 2025-11-21
bump version of cpg flow to incorporate optional assays

Alexander-Stuckey:
# Purpose

Increment the version of `cpg_flow` so that we can skip populating assay information


michael-harper (approved):
Looks good to me

PR #28, opened 2025-11-25
Manifest fix

Alexander-Stuckey:
Specify the manifest file headers in config. Should only need to specify filenames and checksum.


michael-harper (approved):
Config parameters for the win!

PR #38, opened 2026-04-16
Update GVCF blocking strategy and bump to v3.2.1

Alexander-Stuckey:
## Summary
- Update GVCF blocking lower bound from 13 to 10
- Bump version from 3.2.0 to 3.2.1

## Test plan
- [ ] Verify GVCF blocking parameter change produces expected output
- [ ] Confirm version is 3.2.1 across Dockerfile, README, config, and pyproject.toml

### DIFF
--- a/src/dragen_align_pa/jobs/run_align_genotype_with_dragen.py
+++ b/src/dragen_align_pa/jobs/run_align_genotype_with_dragen.py
@@ -98,7 +98,7 @@ def _prepare_fastq_inputs(
             value=(
                 '--qc-coverage-reports-1 cov_report,cov_report '
                 "--qc-coverage-filters-1 'mapq<1,bq<0,mapq<1,bq<0' "
-                "'--vc-gvcf-gq-bands 13 20 30 40 '"
+                "'--vc-gvcf-gq-bands 10 20 30 40 '"
             ),
         ),
     ]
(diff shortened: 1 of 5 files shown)


src/dragen_align_pa/jobs/run_align_genotype_with_dragen.py
```diff
@@ -98,7 +98,7 @@ def _prepare_fastq_inputs(
             value=(
                 '--qc-coverage-reports-1 cov_report,cov_report '
                 "--qc-coverage-filters-1 'mapq<1,bq<0,mapq<1,bq<0' "
-                "'--vc-gvcf-gq-bands 13 20 30 40 '"
+                "'--vc-gvcf-gq-bands 10 20 30 40 '"
```
michael-harper: From Slack:
> the syntax was slightly wrong in the additional args field

@katiedelange Alex mentioned the syntax was wrong, I assume this PR should fix the quote structure here?

```suggestion
                '--vc-gvcf-gq-bands 10 20 30 40 '
```

src/dragen_align_pa/jobs/run_align_genotype_with_dragen.py
```diff
@@ -155,7 +155,7 @@ def _build_cram_specific_inputs(
                     '--qc-coverage-count-soft-clipped-bases true '
                     '--qc-coverage-reports-1 cov_report,cov_report '
                     "--qc-coverage-filters-1 'mapq<1,bq<0,mapq<1,bq<0' "
-                    '--vc-gvcf-gq-bands 13 20 30 40 '
+                    '--vc-gvcf-gq-bands 10 20 30 40 '
```
michael-harper, line 158: Interestingly the quoting issue in the `_prepare_fastq_inputs()` isn't n the CRAM path...so why do we have the old blocking strategy `[10 20 30 40 60 80]` and not `[13 20 30 40]`?

michael-harper (approved):
LGTM

src/dragen_align_pa/jobs/run_align_genotype_with_dragen.py
Alexander-Stuckey: The new blocking strategy was never actually run, due to a syntax error that Dragen silently ignored (I used `_` instead of `-` in the flag)

PR #43, opened 2026-05-07
Fix mlr gvcf header

Alexander-Stuckey:
# Purpose

  - The `##GVCFBlock` header information is missing from the gVCFs produced by Dragen MLR. This PR reintroduces them into the header.

## Proposed Changes

  - Download MLR gVCF from ICA into `tmp` storage.
  - Add new stage to reheader the MLR gVCF and write it to permanent storage.
  - Fix a related bug that was stopping data being downloaded into `tmp` storage. 

## Checklist

- Verified using `fewgenomes-test` that MLR gVCFs now contain `##GVCFBlock` information in the header, at the same location as the base gVCF.

### DIFF
--- a/src/dragen_align_pa/jobs/reheader_mlr_gvcf.py
+++ b/src/dragen_align_pa/jobs/reheader_mlr_gvcf.py
@@ -38,14 +38,7 @@ def reheader_mlr_gvcf(
         f"""
         set -euo pipefail
 
-        bcftools view -h {gvcf_input_group['base_gvcf']} > base_header.txt
-        bcftools view -h {gvcf_input_group['recal_gvcf']} > recal_header.txt
-
-
-        awk 'FNR==NR {{ if (/^##GVCFBlock/) blocks = blocks $0 ORS; next }} \\
-        !inserted && /^##INFO=/ {{ printf "%s", blocks; inserted = 1 }} \\
-        {{ print }}' base_header.txt recal_header.txt > new_header.txt \\
-        && bcftools reheader -h new_header.txt {gvcf_input_group['recal_gvcf']} -o {reheadered_gvcf_outputs['gvcf.gz']}
+        bcftools annotate -h <(bcftools view -h {gvcf_input_group['base_gvcf']} | grep '^##GVCFBlock') {gvcf_input_group['recal_gvcf']} -o {reheadered_gvcf_outputs['gvcf.gz']}
 
         # Index the reheadered gVCF
         bcftools index -t {reheadered_gvcf_outputs['gvcf.gz']}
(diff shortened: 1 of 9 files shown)


src/dragen_align_pa/jobs/reheader_mlr_gvcf.py
```diff
@@ -0,0 +1,57 @@
...
+        set -euo pipefail
+
+        bcftools view -h {gvcf_input_group['base_gvcf']} > base_header.txt
+        bcftools view -h {gvcf_input_group['recal_gvcf']} > recal_header.txt
+
+
+        awk 'FNR==NR {{ if (/^##GVCFBlock/) blocks = blocks $0 ORS; next }} \\
+        !inserted && /^##INFO=/ {{ printf "%s", blocks; inserted = 1 }} \\
+        {{ print }}' base_header.txt recal_header.txt > new_header.txt \\
+        && bcftools reheader -h new_header.txt {gvcf_input_group['recal_gvcf']} -o {reheadered_gvcf_outputs['gvcf.gz']}
```
katiedelange: (I say this as someone who holds a special place in my heart for awk) - is this really the simplest way to do it? 

I believe the awk logic is sound, but can we just use bcftools annotate -h to append the GVCFBlock information to the header directly instead?

```suggestion
        bcftools annotate -h <(bcftools view -h {gvcf_input_group['base_gvcf']} | grep '^##GVCFBlock') {gvcf_input_group['recal_gvcf']} -o {reheadered_gvcf_outputs['gvcf.gz']}
```

katiedelange: Or with a separate line for the bcftools view -h to extract the ##GVCFBlock section first, if that's cleaner.

katiedelange (commented):
One edit proposed - I love awk, but I think it's a bit hard to follow in this context!

katiedelange:
Can we also fix up the pre-commit hooks and lint checker please?

src/dragen_align_pa/jobs/reheader_mlr_gvcf.py
Alexander-Stuckey: To be fair, the only reason I used awk was to put the lines back in the header at exactly the same location as they were in the base gvcf. We totally don't need to do that.

I will swap it to `bcftools annotate`

Alexander-Stuckey:
> Can we also fix up the pre-commit hooks and lint checker please?

All fixed

katiedelange:
Has the bcftools annotate change been re-evaluated on fewgenomes?

katiedelange (approved):
LGTM

PR #50, opened 2026-05-15
Batched DRAGEN submission with per-sample retry

Alexander-Stuckey:
## Summary

PR 2 of 5 for Segment B of the unified DRAGEN pipeline migration. Full rewrite of `manage_dragen_pipeline.run` for batched submission with single-pass per-SG retry (Task 15) plus the deferred Task 14c retry-batch orchestration tests.

The new orchestrator is **not yet wired into any cpg-flow stage** — `stages.py` still references the legacy `run_align_genotype_with_dragen` submitter. Wiring lands in PR 3. The new code is dormant in dev until then.

### What changes

- **`src/dragen_align_pa/jobs/manage_dragen_pipeline.py` (full rewrite, ~440 LOC):**
  - `CohortCancelled` (distinct exception so cpg-flow / operators can tell user-cancellation apart from threshold-breach failures).
  - `_build_submit_callable`: wraps `submit_dragen_batch.run`, threads error_strategy from the batch entry, persists per-SG state FIRST then batches.json SECOND (commit-point ordering — see BatchesFile docstring).
  - `_persist_per_sg_state_for_batch`: writes per-SG state files at the schema version validated by `get_ica_sample_folder`; warns when `expected_outputs` undercounts so the orchestrator surfaces the misconfiguration before submission.
  - `_harvest_ar_guids_from_per_sg_state`: lifts `{batch_index → ar_guid}` out of existing per-SG state files; used by `force_resubmit` before deletion.
  - `_on_succeeded_factory`: transactional SUCCEEDED callback — fetches `passfail.json` in-memory, records per-SG Success/Fail, persists folder FID, marks batch SUCCEEDED. Raises on ICA fetch failure so the loop leaves status INPROGRESS and re-fires next poll (capped by Segment A's on_succeeded failure counter).
  - `_on_status_change_factory`: best-effort mirror of loop-side FAILED_FINAL / CANCELLED transitions into batches.json so resume + retry-building see the latest state.
  - `_build_retry_batches`: collects per-sample failures from `passfail` (or all SGs of a whole-batch FAILED with no passfail.json), forms retry batches via `BatchesFile.add_retry_batch`, marks source-batch SGs retried. CANCELLED is **terminal** — only FAILED feeds the retry path.
  - `_build_loop_outputs_for_batches`: subsets `outputs` keyed by batch name; raises with a friendly batch-size-changed hint if `expected_outputs` is missing entries.
  - `_handle_management_flags`: owns `force_resubmit` (harvest + delete) and `cancel_cohort_run` (cancel in-flight + raise `CohortCancelled` + **preserve per-SG state** for future force_resubmit harvest). Mutual-exclusion checks documented.
  - `_threshold_breached(n_failed, n_total)`: pure 5% strict-greater check, extracted so production code and boundary tests share the comparator.
  - `run()`: full flow — management flags, initial pass, retry pass (handles in-flight resume too), resume-after-cancel guard, 5% threshold with errors.log artefact.

- **`tests/test_manage_dragen_pipeline.py` (new, 16 tests):** retry-batch orchestration (empty/passfail-fail/whole-batch-FAILED/CANCELLED-terminal/single-retry-invariant/audit-trail/find-latest-generation), `_harvest_ar_guids_from_per_sg_state`, `_handle_management_flags` (force_resubmit deletes + harvests, cancel raises + preserves per-SG state), `_threshold_breached` boundary semantics (5%/just-above/small-N).

- **`.pre-commit-config.yaml`:** adds the populationgenomics `cpg-id-checker` hook (parity with main's #42). Test sg_names use `SYN_*` / `f'SYN{i:05d}'` so they pass the hook's `[CX]PG\d+` regex without relying on f-string evasion.

### Compatibility with Segment A changes

The plan code was reviewed against the per-batch-schema, status-enum, and error_strategy validations added in Segment A:

- `record_status` calls use only allowed strings (`SUCCEEDED`, `FAILED`, `CANCELLED`).
- `record_error_strategy` calls use strategies set via `BatchesFile.initialise`/`add_retry_batch` (always in `{auto, continue, terminate}`).
- `_handle_management_flags`'s cancel path uses `record_status(idx, 'CANCELLED')` instead of direct dict mutation, so the value goes through `ALLOWED_BATCH_STATUSES` validation (same outcome today; robust against future enum drift).

### Test plan

- [x] `pytest -q` from this branch: **88 passed** (72 from prior + 16 new).
- [x] `pre-commit run --files <touched>` clean: end-of-file-fixer, trailing-whitespace, ruff v0.15.0, mypy v1.19.1, cpg-id-checker — all pass.
- [x] Test coverage focused on the algorithmic helpers (`_build_retry_batches`, `_handle_management_flags`, `_threshold_breached`). Loop integration glue (`_on_succeeded_factory`, `_on_status_change_factory`, `_build_submit_callable`) is left for end-to-end validation in PR 3 — those helpers' behaviours depend on cpg-flow runtime state that can't reasonably be mocked.
- [ ] **Not exercised:** ICA submission, real GCS reads/writes, real `passfail.json` fetch. Validated end-to-end after PR 3's stage rewiring lands.

### Out of scope (follow-up)

- The new `cpg-id-checker` hook flags pre-existing Segment A docs/READMEs (plan doc, design doc, `README.md`, `README_developer.md` — 25 hits across 4 files). Separate cleanup PR.

### Segment B PR plan

- [#49 — Cache + retry `get_ica_secrets()` (open)](https://github.com/populationgenomics/dragen_align_pa/pull/49)
- **(this PR)** Batched submission orchestrator + retry-batch tests
- Wire stages onto the new orchestrator + add `DownloadBatchArtefactsFromIca`
- Rewire download stages + MLR housekeeping for the new per-SG state shape
- Delete legacy per-sample submitter (`run_align_genotype_with_dragen.py` + `upload_fastq_file_list.py`)


src/dragen_align_pa/jobs/manage_dragen_pipeline.py
```diff
@@ -1,66 +1,845 @@
...
+        if key not in outputs:
+            # `expected_outputs` undercount — downstream per-SG download stages
+            # will fail to resolve the path. Surface this loudly rather than
+            # silently dropping the SG's state file.
+            logger.warning(
+                f'Per-SG state output {key!r} not declared in outputs for batch '
+                f'{batch.name}; downstream download stages for this SG will fail. '
+                f'Check ManageDragenPipeline.expected_outputs.',
+            )
+            continue
```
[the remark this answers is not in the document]
Alexander-Stuckey: Good catch — addressed in [460a93f](https://github.com/populationgenomics/dragen_align_pa/pull/50/commits/460a93f).

Two-layer fix:

1. `run()` now pre-validates that every cohort SG has its `{sg}_pipeline_id_and_arguid` key in `outputs` BEFORE any ICA submission. A missing key raises with the full list of offending SGs, zero ICA work leaked.
2. `_persist_per_sg_state_for_batch` raises (instead of warn+continue) if the case is still reached — defensive code for the cohort-list-changed-mid-run scenario that the pre-check can't have caught.

The "before submission" wording is now accurate.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
```diff
@@ -1,66 +1,845 @@
...
+        # Treating an absent passfail.json as "all SGs Fail" is the spec'd
+        # fallback; the retry pass then resubmits those SGs.
+        if passfail is None:
+            logger.warning(
+                f'Batch {batch.name} (analysis {batch_entry["pipeline_id"]}, '
+                f'folder {analysis_folder_name}): passfail.json not found at ICA root; '
+                f'treating all SGs as Fail.',
+            )
+            batches_file.record_passfail(batch.batch_index, dict.fromkeys(batch.sg_names, 'Fail'))
+        else:
```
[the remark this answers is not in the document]
Alexander-Stuckey: Sharp catch. Flipped the contract in [babe6e5](https://github.com/populationgenomics/dragen_align_pa/pull/50/commits/babe6e5):

- `fetch_passfail_from_ica` now returns `None` **only** when `passfail.json` is legitimately absent (`FileNotFoundError` from the lookup).
- Every other failure (`icasdk.ApiException`, `requests.RequestException`, `requests.HTTPError` from a second 403, `json.JSONDecodeError` from a non-JSON body) is logged and re-raised.

`_on_succeeded` already wraps the fetch in a try/except that translates failures into `RuntimeError`, which the monitor loop handles transactionally (status stays INPROGRESS, next poll re-fires) and the Segment A `MAX_CONSECUTIVE_ON_SUCCEEDED_FAILURES` cap escalates to FAILED_FINAL after persistent failures. So no caller-side change was needed — the new contract slots into the existing transactional flow. Wasted-retry hazard is gone.

The 5 `returns_none_on_*` tests are now `raises_on_*` tests asserting the right exception propagates; happy-path / file-missing / one-retry-403 tests are unchanged.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
```diff
@@ -1,66 +1,845 @@
...
+            is_mlr_pipeline=False,
+            success_file_key_template='{target_name}_success',
+            pipeline_id_file_key_template='{target_name}_pipeline_id',
+            error_log_key=f'{cohort.name}_errors',
+            submit_function_factory=submit_factory,
+            allow_retry=False,
+            sleep_time_seconds=600,
+            on_succeeded=on_succeeded,
+            on_status_change=on_status_change,
+        )
```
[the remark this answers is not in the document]
Alexander-Stuckey: Good catch. Addressed in [a59a094](https://github.com/populationgenomics/dragen_align_pa/pull/50/commits/a59a094): the loop's mid-iteration cancel block now wraps `cancel_ica_pipeline_run.run(...)` in a try/except that logs the failure but still marks the target CANCELLED locally — mirroring `_handle_management_flags`'s pre-loop cancel behaviour.

End-to-end consequence: when `cancel_cohort_run=true` is toggled between polling-loop iterations, an ICA blip during the abort API call no longer propagates out of `manage_ica_pipeline_loop` as a generic Exception. Instead, the loop completes normally, `on_status_change` writes `'CANCELLED'` into `batches.json`, and `run()`'s resume-after-cancel guard raises `CohortCancelled` — uniformly for both pre-loop and mid-loop cancellation timing.

No new unit test added: the polling loop has too many side-effecting deps to mock cleanly, and the new behaviour is symmetric to `_handle_management_flags`'s pre-loop cancel block (already covered by `test_cancel_cohort_run_preserves_per_sg_state`). Happy to extract a testable helper if you'd prefer — would mirror what Segment A did for `_process_succeeded_transition`.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
```diff
@@ -1,66 +1,865 @@
...
+        f'{sg}_pipeline_id_and_arguid'
+        for sg in sg_names
+        if f'{sg}_pipeline_id_and_arguid' not in outputs
+    ]
+    if missing_state_keys:
+        raise KeyError(
+            f'Cohort {cohort.name}: expected_outputs missing per-SG state file entries '
+            f'{missing_state_keys}. Downstream per-SG download stages need these to '
+            f'resolve ICA folder paths. Check ManageDragenPipeline.expected_outputs.',
         )
```
michael-harper, line 620: This feels a little redundant?

michael-harper: Actually, if we change `expected_outputs()` at all this will prevent ICA from firing off and spending $$$

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
```diff
@@ -1,66 +1,865 @@
...
 
-    pipeline_id_key = '{target_name}_pipeline_id_and_arguid'
+    batches_file_path: cpg_utils.Path = outputs[f'{cohort.name}_batches']
```
michael-harper, line 622: This is an example of the 'wiring' not being wired yet? I can't see `outputs[f'{cohort.name}_batches']` in `expected_outputs()` yet.

michael-harper: Nevermind, copilot is on the job!

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
```diff
@@ -1,66 +1,865 @@
...
+    Returns `({batch_index: ar_guid}, {batch_index: {sg_name, ...}})`:
+    - The AR-GUID map is empty unless `force_resubmit` harvested some.
+    - The membership map carries the old per-batch SG set so the caller can
+      warn when the new partition diverges (cohort membership changed).
+
+    Raises `CohortCancelled` (terminal) if `cancel_cohort_run=true` —
+    short-circuits `run()` so it doesn't fall into retry-building or
+    threshold-checking.
+
+    Deliberate spec drift (design doc §4 line 214): the original spec said
```
michael-harper: Do these references to design docs eventually get removed?

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
```diff
@@ -1,66 +1,865 @@
...
+            'force_resubmit and cancel_cohort_run are mutually exclusive. '
+            'To cancel then resubmit: run with cancel_cohort_run=true alone, '
+            'wait for ICA aborts to settle, then rerun with force_resubmit=true.',
+        )
+    if force_resubmit and monitor_previous:
+        logger.warning(
+            f'Cohort {cohort_name}: both force_resubmit and monitor_previous are set; '
+            f'force_resubmit wins and monitor_previous is ignored.',
+        )
+        monitor_previous = False
```
michael-harper: This is a good use case for a config validator in this repo as well. `raise`-ing here may make more sense.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
```diff
@@ -1,66 +1,865 @@
...
+        if had_prior_state:
+            logger.warning(
+                f'force_resubmit=true for cohort {cohort_name}: harvesting AR GUIDs and '
+                f'deleting batches file + per-SG state.',
+            )
+        else:
+            logger.info(
+                f'force_resubmit=true for cohort {cohort_name} but no prior state exists; '
+                f'proceeding as a fresh submission.',
+            )
```
michael-harper, line 512: Silently falling through to a fresh submission when the user is trying to resubmit with no prior state may lead to an expensive ICA run on a cohort the user didn't mean. Could we raise here instead?
`force_resubmit=true requested but no prior state for cohort X; remove the flag to submit fresh`

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
```diff
@@ -1,66 +1,865 @@
...
+        try:
+            with path.open('r') as fh:
+                state = json.load(fh)
+        except (json.JSONDecodeError, OSError) as e:
+            logger.warning(f'Could not harvest AR GUID from {path}: {e}')
+            continue
+        batch_index = state.get('batch_index')
+        ar_guid = state.get('ar_guid')
+        if batch_index is None or not ar_guid:
+            continue
```
michael-harper: `OSError` / `JSONDecodeError` get a warning, but a file that successfully parses to `{}` (which by the looks of it means no batch_index, no ar_guid) just skips silently. Either both should warn, or the parse failure can also be silent. Probably warn-both is the right move if state files are expected to always have these fields.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
```diff
@@ -1,66 +1,865 @@
...
+        # Earlier entries win; subsequent SGs in the same batch should agree.
+        harvested.setdefault(batch_index, ar_guid)
```
michael-harper: This is the same as my comment earlier: if we ran `force_resubmit=true` and the function harvested 0 AR GUIDs from a 100-SG cohort, that's an alarm bell, not a non-event. The caller (`_handle_management_flags`) gets back an empty dict and proceeds.

michael-harper: I think a length check here with at least a warning (if not a `raise`) makes sense.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
```diff
@@ -1,66 +1,865 @@
...
+        else:
+            logger.info(
+                f'force_resubmit=true for cohort {cohort_name} but no prior state exists; '
+                f'proceeding as a fresh submission.',
+            )
+        preserved_ar_guids, old_membership = _harvest_ar_guids_from_per_sg_state(
+            sg_names=sg_names, outputs=outputs,
+        )
+        if batches_file_path.exists():
+            batches_file_path.unlink()
```
michael-harper: Minor, but it looks like this can be one line: `batches_file_path.unlink(missong_ok=True)`

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
```diff
@@ -1,66 +1,865 @@
...
+    if cancel_cohort_run and batches_file_path.exists():
+        logger.warning(f'cancel_cohort_run=true for cohort {cohort_name}: aborting in-flight batches.')
+        existing = BatchesFile(path=batches_file_path)
+        existing.read()
+        n_aborted = 0
+        for b in existing.batches:
+            if b['status'] not in {'PENDING', 'INPROGRESS'}:
+                # Already-SUCCEEDED batches are left alone.
+                continue
+            pipeline_id = b.get('pipeline_id')
```
michael-harper, line 548: I'm just now realising a confusing name-clash with Hail Batch and our 'Batches' concept...

michael-harper: Is there appetite to change this to something like `IcaBatch` or `DragenBatch`?

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
```diff
@@ -1,66 +1,865 @@
...
+        batches_file.read()
+        # Resume uses `retry_generation == 0` (initial batches only) + status to decide
+        # what to re-monitor on the first pass. Retry-batch resumption is handled in the
+        # second loop call below — see retry section.
+        initial_batches = [
+            Batch(cohort_name=cohort.name, batch_index=b['batch_index'], sg_names=b['sg_names'])
+            for b in batches_file.batches
+            if b['retry_generation'] == 0 and b['status'] in {'PENDING', 'INPROGRESS'}
+        ]
+    else:
```
michael-harper, line 652: the post-`force_resubmit` depends on `_handle_management_flags` successfully deleting the batches file. Have we ever experience a failure during deletion that could silently mean `batches_file_path.exists()` is still `True`?

michael-harper: Maybe an `assert not batches_file_path.exists()` at the top of this `else` block would suffice

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
```diff
@@ -1,66 +1,865 @@
...
+                new_set = set(batch_entry['sg_names'])
+                if old_set is not None and old_set != new_set:
+                    added = new_set - old_set
+                    removed = old_set - new_set
+                    logger.warning(
+                        f'force_resubmit: batch {new_index} reuses AR GUID {preserved!r} '
+                        f'but membership has drifted (added={sorted(added)}, '
+                        f'removed={sorted(removed)}). The AR GUID will be associated '
+                        f'with the new membership in this run.',
+                    )
```
michael-harper: This logic is a bit hard to follow as I'm struggling to conceptualise how we keep track of sequencing groups in these runs with the folders on ICA that I think are defined by the `user_reference` naming scheme? However, is it possible that a new sequencing group get's associated with an analysis-runner GUID it was never a part of? What are the implications of this in metamist?

michael-harper (changes requested):
This looks good, few comments and discussion points

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Yeah I will remove them when the actual pipeline code is done, here they (hopefully) just add a bit of context

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Can make that change for sure. I will use `IcaBatch`

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: The main reason for all this ar-guid preservation is that the ar-guid of the original run becomes a part of the ICA folder path for outputs. I would prefer _not_ to do this, but this is the only way to get any kind of billing association between GCP and ICA. We can set the ar-guid as a tag on the pipeline run, but tags are explicitly not returned for any ICA API query, as some users add private information into the tags (convo per Illumina). So we'd have no way to fetch the ar-guid for billing information if we did that.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Agree, will swap to `Raise` here

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Will `Raise` here as well

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: I think a `Raise` on `{}` also makes sense. This shouldn't have happened, so we should error and find out why.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Agree with `Raise`

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Good catch

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Can do

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
michael-harper: It's a nuisance but I understand the necessity of recording ARGUIDS. I'm asking whether preserving them like this could cause batch membership to drift. 
I Claude to mocked up an example and it provided some more insights:
"""
Concrete case: `batch 0` originally ran with `{A, B, C, D, E}` under `AR GUID X`. User changes the cohort (drops `C`, adds `F`), then `force_resubmit=true`. The new `batch 0` contains `{A, B, D, E, F}` and inherits `AR GUID X` — including for SG `F`, which was never part of `AR GUID X`'s original run. The per-SG state file for `F` now points at `AR GUID X`. If Metamist indexes "which AR GUIDs has this SG been part of," `F` shows up linked to an AR GUID it didn't actually run under. The billing trail for the SGs that were in the original batch is still intact — but newcomer SGs get a misleading association.
"""

It also asked a reasonable question "Would it be safer to skip the AR GUID pre-seed when old_set != new_set and mint a fresh GUID for batches with drifted membership?" but I don't think we can generate AR GUID's on the fly?

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Done in [439f873](https://github.com/populationgenomics/dragen_align_pa/commit/439f873), further simplified in [4e68b38](https://github.com/populationgenomics/dragen_align_pa/commit/4e68b38) — `JSONDecodeError` / `OSError` now propagate naturally; a parsed-but-empty state file raises `KeyError` naming the missing field.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Done in [439f873](https://github.com/populationgenomics/dragen_align_pa/commit/439f873); the zero-harvest `RuntimeError` was simplified away in [4e68b38](https://github.com/populationgenomics/dragen_align_pa/commit/4e68b38) — any unparseable state file now raises `JSONDecodeError` / `OSError` naturally, which trips the same alarm bell.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Done in [439f873](https://github.com/populationgenomics/dragen_align_pa/commit/439f873) — `force_resubmit` + `monitor_previous` now raises `ValueError(... mutually exclusive ...)`, symmetric with the existing `force_resubmit` + `cancel_cohort_run` check immediately above.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Done in [439f873](https://github.com/populationgenomics/dragen_align_pa/commit/439f873) — `force_resubmit=true` with no prior state now raises `RuntimeError` with the message you suggested.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Done in [439f873](https://github.com/populationgenomics/dragen_align_pa/commit/439f873).

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Done in [7dea97c](https://github.com/populationgenomics/dragen_align_pa/commit/7dea97c) — renamed everywhere except in user-facing log strings (`'Batch COH0001-batch0000 …'`) where the cohort-prefixed identifier is already unambiguous and `IcaBatch` reads worse.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Done in [439f873](https://github.com/populationgenomics/dragen_align_pa/commit/439f873) — `assert not batches_file_path.exists()` at the top of the post-`force_resubmit` `else` block. (`.unlink(missing_ok=True)` upstream of it makes the deletion itself loud-on-failure too.)

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Hmm, I can agree that upon a `force_resubmit`, the user's intent is to start with a clean slate, so `au-guid` and previous batch state shouldn't be preserved.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
```diff
@@ -1,66 +1,845 @@
...
-    """
-    Calls the generic pipeline manager with settings for the main Dragen pipeline.
-    """
+    """Build batches, submit them, retry per-sample failures once, enforce 5% threshold."""
+    batch_size: int = config_retrieve(['dragen_align_pa', 'manage_dragen_pipeline', 'batch_size'], default=5)
+    sg_names = [sg.name for sg in cohort.get_sequencing_groups()]
+    if not sg_names:
+        raise ValueError(f'Cohort {cohort.name} has no sequencing groups.')
+    batches_file_path: cpg_utils.Path = outputs[f'{cohort.name}_batches']
+
```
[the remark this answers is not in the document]
Alexander-Stuckey: The signature mismatch is real but intentional: this PR is staged with PR #3 (which rewires the `ManageDragenPipeline` stage to the new `run()` shape and the new `expected_outputs` keys). `dragen-unified-dev` sees the two PRs merged together; `main` is untouched until the migration is complete. No backward-compatible shim — the legacy submitter is being removed wholesale in the same series.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Agreed — the pre-validation check is intentionally protective. It catches an `expected_outputs` mistake at start-of-run, before any ICA submissions go out, so a typo or stale config can't burn money on orphaned analyses with no corresponding state files.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Done in [e8cdf9e](https://github.com/populationgenomics/dragen_align_pa/commit/e8cdf9e) — `spec §X line Y` / `Task N` / `design doc §N` citations stripped across 5 files. The explanations the citations decorated are kept; only the cross-doc back-references are gone.

src/dragen_align_pa/jobs/manage_dragen_pipeline.py
Alexander-Stuckey: Implemented in [9af79c3](https://github.com/populationgenomics/dragen_align_pa/commit/9af79c3) — `force_resubmit=true` now deletes the batches file, completion marker, and every per-SG state file, then re-batches from scratch with fresh AR GUIDs. `_harvest_ar_guids_from_per_sg_state` and the pre-seeding block in `run()` are gone (−172 lines net); `ar_guid_override` dropped from `submit_dragen_batch.run`'s signature too. The membership-drift scenario you sketched can no longer happen.

michael-harper: Nice!

michael-harper (approved):
LGTM

PR #52, opened 2026-05-18
Rewire cpg-flow stages onto the batched DRAGEN layout

Alexander-Stuckey:
## Summary

- Migrate the `Download*FromIca` + `ManageDragenMlr` stages to resolve ICA paths via the per-SG state file written by the new orchestrator (extended schema introduced in PR #50). Adds `DownloadBatchArtefactsFromIca` for per-batch passfail / summary / reports/ artefacts; removes obsolete `UploadFastqFileList` (per-batch FASTQ upload folded into the submitter).
- Extracts shared helpers — `load_per_sg_state`, `get_per_sg_state_path`, `make_download_job`, `resolve_and_run` — and collapses three near-identical Download stage bodies onto them.
- Three rounds of simplification on top of the functional work: idiomatic Python, comment trimming, log-noise cleanup. **−224 lines net** (1114+ / 487−) on top of feature work, no behavior change.

Closes plan Tasks 16, 17, 18, 19, 19b, 20, 21, 22. `DeleteDataInIca` rewrite and Tasks 23–24 (legacy entrypoint delete + README/DAG refresh) deferred to follow-up PRs per the plan.

## Test plan

- [x] `pre-commit run` — pinned ruff v0.15.0 + mypy v1.19.1 pass on every changed file
- [x] `pytest -q` — 122 tests pass (new unit tests for `DownloadBatchArtefactsFromIca`, `prepare_ica_for_analysis`, and the aggregated missing-keys behavior of `load_per_sg_state`)
- [x] `pytest --collect-only` — full stage graph imports cleanly
- [ ] End-to-end cohort run against ICA — Task 25 post-merge validation checklist (run after this lands on `dragen-unified-dev`)


src/dragen_align_pa/jobs/prepare_ica_for_analysis.py
```diff
@@ -10,44 +10,48 @@
...
-    Args:
-        cohort (Cohort): The Cohort being aligned and genotyped
-        outputs (dict): Dictionary specifying ICA output folder IDs per sequencing group.
+def run(cohort: Cohort, output: cpg_utils.Path) -> None:
+    """Create (or find) a single ICA folder for the cohort's pipeline outputs.
 
-    Returns:
-        dict [str, str] noting the analysis ID.
+    The new unified pipeline writes per-batch analysis folders directly under
+    this parent folder, so we no longer need per-SG parent folders.
```
michael-harper: I would write these comments so as to avoid reference to any 'old' way. 
"This pipeline writes per-batch (not per-SG) analysis folders directly under this parent folder."

src/dragen_align_pa/jobs/prepare_ica_for_analysis.py
```diff
@@ -10,44 +10,48 @@
...
-            with outputs[sg_name].open('w') as opath:
-                opath.write(json.dumps({'analysis_output_fid': object_id}))
+        folder_id, status = ica_utils.create_upload_object_id(
+            api_instance=api_instance,
+            path_params={'projectId': project_id},
+            sg_name=cohort.name,
+            file_name=cohort.name,
+            folder_path=folder_path,
+            object_type='FOLDER',
+        )
```
michael-harper, line 38: We should change `sg_name` here. It only ever get's passed a `cohort.name`. Suggesting something like `object_name` or `target_name`. 

I think this can actually safely just be called `folder_name` as per Claude's assessment:
```
 The three sites all pass cohort-level names:

- prepare_ica_for_analysis.py:31 → sg_name=cohort.name
- submit_dragen_batch.py:291 → sg_name=cohort_name (and object_type is 'FILE', where sg_name is ignored entirely — see line 65–70: the FILE branch uses file_name, not sg_name)
- manage_md5_pipeline.py:70 → sg_name=cohort_name
```

src/dragen_align_pa/ica_utils.py
```diff
@@ -314,7 +312,6 @@ def finalize_upload(
     Re-fetches the file ID from ICA and writes the output JSON file.
```
michael-harper, line 312: Can we change this functions name to `finalise_upload`?

src/dragen_align_pa/stages.py
```diff
@@ -262,97 +245,56 @@ def queue_jobs(self, sequencing_group: SequencingGroup, inputs: StageInput) -> S
...
-
-    This stage handles the following tasks:
-    1. Cancels a previous pipeline running on ICA if requested.
-        - Set the `cancel_cohort_run` flag to `true` in the config and the stage will read the pipeline ID from the JSON file and cancel it.
-    2. Resumes monitoring a previous pipeline run if it was interrupted.
-        - Set the `monitor_previous` flag to `true` in the config. This will read the pipeline ID from the JSON file and monitor it.
-    3. Initiates a new Dragen pipeline run if no previous run is found or if resuming is not requested.
-    4. Monitors the progress of the Dragen pipeline run.
-    """  # noqa: E501
+    """Submit cohort batches to the unified DRAGEN pipeline and monitor them."""
```
michael-harper:
```suggestion
class ManageDragenPipeline(CohortStage):
    """Submit cohort batches to the unified DRAGEN pipeline and monitor them.

    Inputs (selected by READS_TYPE):
        cram: per-SG ICA state from UploadDataToIca
        fastq: fastq_ids map from FastqIntakeQc + per-SG CSVs from MakeFastqFileList
                (combined into per-batch CSVs and uploaded to ICA inline by the submitter)
        + analysis output folder FID from PrepareIcaForDragenAnalysis (both modes)

    Outputs (2 cohort-level entries + one per sequencing group):

        {
            '<cohort>_batches':            <cohort>_batches.json            # batch plan
            '<cohort>_pipeline_complete':  <cohort>_pipeline_complete.json  # success marker
            '<sg>_pipeline_id_and_arguid': <sg>_pipeline_id_and_arguid.json # per-SG ICA state
            ... one of these per SG ...
        }

    All paths live under `get_pipeline_path()`. See `expected_outputs` for why
    per-batch scratch files (errors.log, per-batch success/pipeline_id) are
    deliberately excluded from this dict.
    """

```

src/dragen_align_pa/utils.py
```diff
@@ -152,56 +154,36 @@ def get_batch_artefacts_path(cohort_name: str, batch_index: int) -> cpg_utils.Pa
...
+    sequencing_group: SequencingGroup,
+    state_stage: Callable[..., Stage],
+) -> tuple[Cohort, cpg_utils.Path]:
+    """Look up an SG's per-SG state file via the given upstream stage's outputs.
+
+    Returns `(cohort, state_path)` because both are needed at every call site:
+    the cohort to read `inputs.as_dict`, and `cohort.name` to thread into
+    downstream `resolve_and_run` / `get_ica_sample_folder` calls.
+    """
+    cohort = get_multicohort().get_cohorts()[0]
```
michael-harper: Would we ever be running MultiCohort's through this pipeline? Indexing `[0]` would mean we get every SG for the first cohort only

src/dragen_align_pa/jobs/download_batch_artefacts.py
```diff
@@ -0,0 +1,263 @@
...
+                    api_instance=api_instance,
+                    path_parameters=path_parameters,
+                    base_ica_folder_path=reports_folder,
+                    recursive=True,
+                )
+            except icasdk.ApiException as e:
+                logger.warning(
+                    f'Batch {batch_name}: reports/ folder not enumerable at '
+                    f'{reports_folder} ({e}); skipping.',
+                )
```
michael-harper, line 247: This exception doesn't increment `stats.lookup_failure`. We should probably increment. Or, if we want to specifically address the `reports/` folder not being enumerable we could narrow the exception (if possible) and let other API errors propagate.

michael-harper (changes requested):
Downloading looks much more streamlined. Few comments, I don't think anything major.

src/dragen_align_pa/jobs/prepare_ica_for_analysis.py
Alexander-Stuckey: Done in 5e831b8 — reworded to "This pipeline writes per-batch (not per-SG) analysis folders directly under this parent folder." (verbatim from your suggestion).

src/dragen_align_pa/jobs/prepare_ica_for_analysis.py
Alexander-Stuckey: Done in 5e831b8 — renamed parameter to `folder_name` in `create_upload_object_id` (only used when `object_type == "FOLDER"`, ignored otherwise) and updated all three call sites (`prepare_ica_for_analysis`, `submit_dragen_batch`, `manage_md5_pipeline`).

src/dragen_align_pa/ica_utils.py
Alexander-Stuckey: Done in 5e831b8 — `finalize_upload` → `finalise_upload`, plus the one caller in `upload_data_to_ica.py`.

src/dragen_align_pa/stages.py
Alexander-Stuckey: Applied the suggested docstring verbatim in 5e831b8.

src/dragen_align_pa/utils.py
Alexander-Stuckey: Good catch. Fixed in 5e831b8: instead of `get_cohorts()[0]`, the helper now finds the cohort that actually contains the SG and raises a `ValueError` if 0 or >1 cohorts match. Works correctly under both single- and multi-cohort runs, and fails loudly rather than silently picking the wrong cohort if anything is misconfigured.

src/dragen_align_pa/jobs/download_batch_artefacts.py
Alexander-Stuckey: Addressed in 5e831b8 by combining both options: narrowed to `HTTPStatus.NOT_FOUND` (expected — `reports/` is genuinely absent on a catastrophically-failed batch, silently skipped); any other `ApiException` now increments `stats.lookup_failure` and logs, so it surfaces in the run summary. Did not re-raise non-404s — that would let a single transient 5xx on one batch's `reports/` lookup kill the whole cohort's artefact download, even when other batches succeeded.

src/dragen_align_pa/utils.py
michael-harper: Claude reckons: same SG is represented by two distinct Python objects in two cohort objects — without it, you'd silently get len(matching) == 0 and raise even though the SG logically belongs to both.
```suggestion
    matching = [c for c in get_multicohort().get_cohorts() if sequencing_group.id in c.get_sequencing_groups_ids()]
```

michael-harper: I think my original comment is a moot point now that I look back at it as we won't run multicohorts through this pipeline. Indeed catching this duplication here would be too late anyway. `ManageDragenPipeline` is a `CohortStage`, so cpg-flow runs it once per cohort and the SG would have already been submitted to ICA multiple times (once per cohort) and the money spent. 

If we were to ever support multicohorts with overlapping SGs, the guard against duplicate submissions should be in `ManageDragenPipeline` I believe.

michael-harper (changes requested):

src/dragen_align_pa/jobs/download_batch_artefacts.py
```diff
@@ -0,0 +1,271 @@
...
+    from a transfer/GCS outage (stream-heavy). `FileNotFoundError` at lookup
+    is legitimate absence and is NOT counted as a failure."""
+
+    success: int = 0
+    lookup_failure: int = 0
+    stream_failure: int = 0
+
+    @property
+    def total_failure(self) -> int:
+        return self.lookup_failure + self.stream_failure
```
michael-harper:
```suggestion
from dataclasses import dataclass, field

@dataclass
class _StreamStats:
    """Per-cohort marker-payload tracking.

    Tracks *identities* (not just counts) for failures so an operator reading
    the marker payload can see which specific files failed and run a targeted
    retry instead of having to scrape job logs.

    `lookup_failure` and `stream_failure` are split so a cohort-wide auth /
    connectivity outage (lookup-heavy) is distinguishable from a transfer /
    GCS outage (stream-heavy). `FileNotFoundError` at lookup is legitimate
    absence and is NOT recorded as a failure. `skipped` counts files already
    present in GCS from a previous (partial) run.
    """

    success: int = 0
    skipped: int = 0
    lookup_failures: list[dict[str, str]] = field(default_factory=list)
    stream_failures: list[dict[str, str]] = field(default_factory=list)

    @property
    def total_failure(self) -> int:
        return len(self.lookup_failures) + len(self.stream_failures)
```

src/dragen_align_pa/jobs/download_batch_artefacts.py
```diff
@@ -0,0 +1,271 @@
...
+            file_id=file_id,
+            file_name=file_name,
+            gcs_bucket=gcs_bucket,
+            gcs_prefix=gcs_prefix,
+            expected_md5_hash=None,
+        )
+    except (icasdk.ApiException, requests.RequestException, gcs_exceptions.GoogleCloudError) as e:
+        logger.warning(f'{context}: streaming {file_name} failed ({e}); skipping.')
+        stats.stream_failure += 1
+        return
```
michael-harper, line 99:
```suggestion
    except (icasdk.ApiException, requests.RequestException, gcs_exceptions.GoogleCloudError) as e:
        logger.warning(f'{context}: streaming {file_name} failed ({e}); skipping.')
        stats.stream_failures.append({
            'context': context,
            'file_name': file_name,
            'error': str(e),
        })
        return
```

src/dragen_align_pa/jobs/download_batch_artefacts.py
```diff
@@ -0,0 +1,271 @@
...
+            file_name=file_name,
+        )
+    except FileNotFoundError:
+        logger.warning(f'{file_name} not present in {parent_folder}; skipping.')
+        return
+    except icasdk.ApiException as e:
+        logger.warning(
+            f'ICA API error while looking up {file_name} in {parent_folder}: {e}; skipping.',
+        )
+        stats.lookup_failure += 1
```
michael-harper:
```suggestion
        stats.lookup_failures.append({
            'parent_folder': parent_folder,
            'file_name': file_name,
            'error': str(e),
        })
```

src/dragen_align_pa/jobs/download_batch_artefacts.py
```diff
@@ -0,0 +1,271 @@
...
+                        f'Batch {batch_name}: reports/ folder absent at '
+                        f'{reports_folder}; skipping.',
+                    )
+                else:
+                    stats.lookup_failure += 1
+                    logger.warning(
+                        f'Batch {batch_name}: reports/ folder not enumerable at '
+                        f'{reports_folder} ({e}); skipping.',
+                    )
+                report_files = []
```
michael-harper, line 253:
```suggestion
            except icasdk.ApiException as e:
                logger.warning(
                    f'Batch {batch_name}: reports/ folder not enumerable at '
                    f'{reports_folder} ({e}); skipping.',
                )
                stats.lookup_failures.append({
                    'parent_folder': reports_folder,
                    'file_name': '<reports/ enumeration>',
                    'error': str(e),
                })
                report_files = []
```

src/dragen_align_pa/jobs/download_batch_artefacts.py
```diff
@@ -0,0 +1,271 @@
...
+        json.dump(
+            {
+                'cohort_name': cohort_name,
+                'batches_processed': batches_processed,
+                'success_count': stats.success,
+                'lookup_failure_count': stats.lookup_failure,
+                'stream_failure_count': stats.stream_failure,
+            },
+            fh,
+        )
```
michael-harper, line 301:
```suggestion
    with marker_path.open('w') as fh:
        json.dump(
            {
                'cohort_name': cohort_name,
                'batches_processed': batches_processed,
                'success_count': stats.success,
                'skipped_count': stats.skipped,
                'lookup_failures': stats.lookup_failures,
                'stream_failures': stats.stream_failures,
            },
            fh,
            indent=2,
        )
```

src/dragen_align_pa/jobs/download_batch_artefacts.py
```diff
@@ -0,0 +1,271 @@
...
+            f'persists, investigate ICA / GCS connectivity for this project.',
+        )
+
+    summary_msg = (
+        f'Cohort {cohort_name} batch-artefact download: '
+        f'batches_processed={batches_processed}, '
+        f'success={stats.success}, '
+        f'lookup_failures={stats.lookup_failure}, '
+        f'stream_failures={stats.stream_failure}.'
+    )
```
michael-harper, line 283:
```suggestion
    summary_msg = (
        f'Cohort {cohort_name} batch-artefact download: '
        f'batches_processed={batches_processed}, '
        f'success={stats.success}, '
        f'skipped={stats.skipped}, '
        f'lookup_failures={len(stats.lookup_failures)}, '
        f'stream_failures={len(stats.stream_failures)}.'
    )
```

src/dragen_align_pa/jobs/download_batch_artefacts.py
```diff
@@ -0,0 +1,271 @@
...
+    context: str,
+    stats: _StreamStats,
+) -> None:
+    """Stream one file to GCS; warn-and-skip on transient ICA / HTTP / GCS errors.
+
+    Wraps `stream_ica_file_to_gcs` so one transient blip doesn't abort the
+    whole cohort mid-loop. MD5-mismatch (`ValueError`) is intentionally not
+    caught — we pass `expected_md5_hash=None` here so it's unreachable, but a
+    future caller passing a hash should crash hard on integrity violations.
+    """
```
michael-harper, line 77:
```suggestion
def _stream_silently(
    api_instance: project_data_api.ProjectDataApi,
    path_parameters: dict[str, str],
    file_id: str,
    file_name: str,
    gcs_bucket: storage.Bucket,
    gcs_prefix: str,
    context: str,
    stats: _StreamStats,
    ) -> None:
        """Stream one file to GCS; warn-and-skip on transient ICA / HTTP / GCS errors.

        Skips outright if the GCS blob already exists, supporting a 'delete the
        marker and re-run' retry workflow that re-attempts only previously-failed
        files. GCS resumable uploads only finalize on completion, so a stream
        failure in a previous run won't leave a partial blob — an existing blob
        means a previous run committed successfully.

        Wraps `stream_ica_file_to_gcs` so one transient blip doesn't abort the
        whole cohort mid-loop. MD5-mismatch (`ValueError`) is intentionally not
        caught — we pass `expected_md5_hash=None` here so it's unreachable, but a
        future caller passing a hash should crash hard on integrity violations.
        If MD5 verification is ever wired through this function, the skip-if-exists
        branch above needs to be gated on a "verified" sidecar or removed —
        trusting an existing blob is only safe in the no-hash regime.
        """
        gcs_blob_path = f'{gcs_prefix}/{file_name}'
        if gcs_bucket.blob(gcs_blob_path).exists():
            stats.skipped += 1
            return
```

michael-harper (commented):
Going back over this I felt we can do a better job at tracking missed files during the batch artefact download. We were skipping them without logging which ones failed and without any ability to retry partial stage completion. The below is a, hopefully, complete implementation. It's a well contained adjustment that fits inside `download_batch_artefacts.py` nonetheless sorry for the bulk dump at the last minute!

michael-harper (approved):
LGTM!

PR #51, opened 2026-05-15
Populate BED file ICA IDs in constants

Alexander-Stuckey:
## Summary

Replace the four `fil.TODO_REPLACE_AFTER_ICA_UPLOAD` placeholders in `dragen_align_pa.constants.ICA_FILE_IDS` with the real ICA file IDs returned after the BED files were uploaded.

| BED basename | ICA file ID |
|---|---|
| `Twist_VCGS_Exome_Covered_Targets_hg38.bed` | `fil.60130ada16264ed28a7008deb1d54636` |
| `S06588914_Regions_hg38.bed` | `fil.844dfe0ff4c74b10fbc008deabc38c7f` |
| `S30409818_Regions.bed` | `fil.5d4da6b9c2c74abcb00608deb2229b88` |
| `S30409818_Covered.bed` | `fil.625777f457c84b508a7108deb1d54636` |

Stock-config submissions that reference these names (e.g. `coverage_region_beds` in `dragen_align_pa_defaults.toml`) now succeed instead of failing at `resolve_ica_file_id` with the placeholder-rejection error introduced alongside the registry in PR #47.

## Test plan

- [x] `pytest tests/test_constants.py -q`: 3 passed (existing tests monkeypatch the registry, so changing the production values doesn't affect them).
- [x] `pre-commit run --files <touched>` clean.
- [ ] **Not exercised:** end-to-end ICA submission against the real file IDs — validated once Segment B PR 3 wires the new orchestrator into stages.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #55, opened 2026-05-21
Resolve exome BEDs by name; validate cohort design

michael-harper:
## Summary

Two improvements to how this pipeline handles exome BEDs.

**Operators specify exome target BEDs by name, not file ID.** Previously the WES preset shipped `additional_args` strings with `<bed-name>` placeholders that had to be hand-replaced, and `additional_files` took raw `fil.…` IDs that had to be looked up and pasted in separately. Now the run config sets BED basenames once under `[dragen_align_pa.manage_dragen_pipeline.presets.exome.bed_names]` with three fields (`vc_target` / `cnv_target` / `sv_call_regions`), and `submit_dragen_batch.py` i.e. the submitter:

- substitutes them into `additional_args` (now using `{vc_target}` etc. tokens),
- resolves them through `ICA_FILE_IDS` and adds the file IDs to `additional_files` automatically.

This matches the pattern already used for `coverage_region_beds`.

**Exome cohorts get a design check at startup.** A new step inside `PrepareIcaForDragenAnalysis.queue_jobs` reads every SG's `assay.meta['sequencing_library']` from metamist, maps each lab/protocol variant (e.g. `SSQXTCREV2`, `SSXTLICREV2`, `AgilentCREv2WES`) to a canonical design (CRE / CREv2 / Twist) via `DESIGN_TO_CANONICAL`, and:

- fails if the cohort contains a mix of designs (e.g. CRE and Twist SGs in one cohort),
- fails if any BED named in `bed_names` doesn't belong to the cohort's resolved design,
- does nothing for genome runs.

`DESIGN_TO_BEDS` in `constants.py` is the source of truth for which BED basenames belong to which design.

## Depends on

- PR #51 must merge first. `ICA_FILE_IDS` still holds `_TODO_FID` placeholders for the four registered BED basenames; until PR #51 replaces them with real `fil.…` IDs, any run using this code halts at `resolve_ica_file_id`.

## Known gaps

- CRE Covered BED isn't uploaded yet — when a CRE cohort is run, the operator's config must use the Regions BED for all three entries until Covered is registered.
- CNV BED choice for Agilent CREv2 is provisional; the convention used for upcoming runs (Covered for `vc_target`, Regions for `cnv_target` / `sv_call_regions`) is pending next sprint's review.

## Deferred

- `--vc-target-bed-padding` flag. @Alexander-Stuckey: do you know the right DRAGEN arg(s) and default value. Will land on this branch if it arrives before merge, otherwise as a follow-up PR.

## Supersedes

- PR #44 — same intent (per-cohort design validation) but PR44 targeted the old per-SG architecture and the old `[ica.reference_data].exome_bed` config key. This PR re-implements it for the unified batched pipeline.

## Test plan

- [ ] `pre-commit run --files <touched>` — to be run before merge.
- [ ] `pytest tests/test_submit_dragen_batch.py tests/test_utils.py -q` — 15 new tests; not yet run locally.
- [ ] End-to-end submission against ICA.


config/dragen_align_pa_defaults.toml
```diff
@@ -90,12 +90,21 @@ cnv_segmentation_mode = "SLM"
 additional_args = "--cnv-enable-self-normalization true"
 additional_files = []
 
-# WES-only flags. Fill in WES BED + PoN file IDs per cohort.
+# WES-only flags. Your run's config TOML must set [bed_names] below; blank
+# values here are fail-fast markers that halt the run before ICA submission.
 [dragen_align_pa.manage_dragen_pipeline.presets.exome]
 cnv_segmentation_mode = "HSLM"
-additional_args = "--sv-exome true --sv-call-regions-bed <bed-name> --vc-target-bed <bed-name> --cnv-target-bed <bed-name> --cnv-target-factor-threshold 5 --cnv-enable-self-normalization false"
+additional_args = "--sv-exome true --sv-call-regions-bed {sv_call_regions} --vc-target-bed {vc_target} --cnv-target-bed {cnv_target} --cnv-target-factor-threshold 5 --cnv-enable-self-normalization false"
```
Alexander-Stuckey: `--vc-target-padding` can be left off, It just adds extra bases around the supplied capture regions that are allowed for variant calling, which I don't believe we want.
We do need to add `--cnv-merge-threshold 0.4`, which is the default for exome calling (it defaults to 0.2 for genome calling)

src/dragen_align_pa/jobs/submit_dragen_batch.py
```diff
@@ -54,14 +54,35 @@
 _MAX_COVERAGE_REGION_BEDS = 3
 
 
+def _get_bed_names_for_seqtype() -> dict[str, str]:
+    """Read `[presets.<seqtype>.bed_names]` and return its `{key: basename}` map.
+
+    Empty-string values raise — defaults.toml ships them empty as fail-fast
+    markers so an un-overridden run halts before ICA submission.
```
Alexander-Stuckey: This is a bit awkward to read, so I suggest
```
    Empty string values in the config (e.g. vc_target = "") raise ValueErrors.
    This is to fail-fast before any ICA submission. The user must set all presets
    in the config deliberately to avoid this.
```

src/dragen_align_pa/jobs/submit_dragen_batch.py
```diff
@@ -54,14 +54,35 @@
...
+    """
+    sequencing_type = config_retrieve(['workflow', 'sequencing_type'])
+    bed_names = config_retrieve(
+        ['dragen_align_pa', 'manage_dragen_pipeline', 'presets', sequencing_type, 'bed_names'],
+        default={},
+    )
+    if not bed_names:
```
Alexander-Stuckey: I would move the `KeyError` from `_build_additional_args` here, rather than return an empty dict and then raising.

src/dragen_align_pa/jobs/submit_dragen_batch.py
```diff
@@ -91,11 +112,21 @@ def _build_additional_args() -> str:
...
+    bed_names = _get_bed_names_for_seqtype()
+    try:
+        preset_args = preset_args.format(**bed_names)
+        user_args = user_args.format(**bed_names)
+    except KeyError as e:
+        raise ValueError(
+            f'DRAGEN additional_args references {{{e.args[0]}}} but '
+            f'[presets.{sequencing_type}.bed_names] has no entry named '
+            f'{e.args[0]!r}. Configured entries: {sorted(bed_names)}.',
+        ) from None
```
Alexander-Stuckey, line 103: Move to above function as mentioned

src/dragen_align_pa/constants.py
```diff
@@ -39,6 +39,34 @@
...
+    'TwistWES1VCGS1': CANONICAL_DESIGN_TWIST,
+    'AgilentCREv2WES': CANONICAL_DESIGN_CREV2,
+}
+
+# Per canonical design: the BED basenames known to belong to it. The validator
+# in utils.assert_cohort_design_matches_configured_bed uses this to check that
+# the basenames named in [presets.exome.bed_names] match the cohort's resolved
+# design. All entries must also be registered in ICA_FILE_IDS above.
+DESIGN_TO_BEDS: Final[dict[str, frozenset[str]]] = {
+    CANONICAL_DESIGN_CRE: frozenset({'S06588914_Regions_hg38.bed'}),
```
Alexander-Stuckey: This one can be removed, this is the hg19 liftover I believe.

src/dragen_align_pa/jobs/submit_dragen_batch.py
```diff
@@ -54,14 +54,35 @@
 _MAX_COVERAGE_REGION_BEDS = 3
 
 
+def _get_bed_names_for_seqtype() -> dict[str, str]:
```
Alexander-Stuckey: I suggest that you move this entire function into utils.py, as `assert_cohort_design_matches_configured_bed` performs the same function, but with less strictness (this function lists what keys are unset). Then, the `assert_...` function can just call this.

src/dragen_align_pa/utils.py
```diff
@@ -99,6 +104,107 @@ def run_subprocess_with_log(
...
+    )
+    used_basenames = [
+        str(name).strip() for name in bed_names.values() if str(name).strip()
+    ]
+    if not used_basenames:
+        raise RuntimeError(
+            f'Cohort {cohort.id} resolves to design {cohort_design!r} but '
+            f'[presets.exome.bed_names] has no non-blank entries. Set the BED '
+            f'basenames in your run config.',
+        )
```
Alexander-Stuckey: This can be replaced with a call to `_get_bed_names_for_seqtype`, once that function is moved to `utils.py`

Alexander-Stuckey (changes requested):
Looks pretty good, one larger function move to consider.
On moving the function, you'd need to update the tests as well.
Also worth adding a test that checks for partially empty BED names, since that is what this function move is designed to catch.

src/dragen_align_pa/constants.py
```diff
@@ -47,7 +46,7 @@
 
 # Exact-match map from metamist assay.meta['sequencing_library'] values to
 # canonical designs. Populate as new values are encountered in metamist. Exact
-# match (rather than substring) avoids prefix collisions like SSQXTCRE ⊂ SSQXTCREV2.
+# match (not substring) avoids prefix collisions like SSQXTCRE being a prefix of SSQXTCREV2.
 DESIGN_TO_CANONICAL: Final[dict[str, str]] = {
     'SSQXTCRE': CANONICAL_DESIGN_CRE,
```
Alexander-Stuckey: `CANONICAL_DESIGN_CRE` can be removed everywhere 😄

config/dragen_align_pa_defaults.toml
michael-harper: I have added `--cnv-merge-threshold 0.4` to `additional_args`. 

@joshuamschmidt can you provide some clarification over the necessity of `--vc-target-padding`?

joshuamschmidt: @michael-harper @Alexander-Stuckey reposting same as on slack:

A few thoughts :

- build in flexibility by including the padding as configurable, default = 0. This also means we don't need to mess around creating padded bed files if we ever change our mind.

- We will likely use Agilent Covered bedfile with 0bp padding -> this bedfile is defined as 'the space of the probes used to capture the desired targets', and is already broader than the target co-ordinates.

- For the Twist cohort, twist only provides the 'narrower' target coordinates. They say b/c of how these are manufactured. But it is only 40MB versus Agilents ~80MB or so. So i see no harm in adding 50 or 100 bp padding for this cohort.


This is also predicated that hail/gnomad Methods handles reference block qc filtering to get accurate AN for AF calculation.

and we need to be specific in which bed file we pass b/c of the CNV calling step (rather than passing in some union bed file etc)

michael-harper: So we can have three extra fields in the config that specify the padding of each call type:
```TOML
[dragen_align_pa.manage_dragen_pipeline.presets.exome.bed_padding]
vc_target = 0
cnv_target = 0
sv_call_regions = 0
```

@Alexander-Stuckey do you know the naming of the padding flag? (Assuming they have separate padding flags for each type)

Alexander-Stuckey: @michael-harper The flag is `--vc-target-bed-padding` and only applies to the small variant caller, not the SV or CNV caller.

Alexander-Stuckey (approved):
Looks good to me

PR #58, opened 2026-05-26
ManageDragen*: route per-SG paths via tmp manifest + pin non-preempt

Alexander-Stuckey:
## Summary

`ManageDragenPipeline` and `ManageDragenMlr` pass per-SG path dicts (`outputs`, `cram_ica_fids_path`, `analysis_output_fids_path`, `pipeline_id_arguid_path_dict`, fastq variants) inline through `job.call(...)`. Hail Batch embeds path args inline in each PythonJob spec, which is capped at 1 MiB — the spec grows linearly with cohort size and overflows. The dispatcher now writes one flat JSON manifest to a tmp GCS path and threads only the manifest path through `job.call`; the worker reads it at runtime. Spec growth becomes O(1).

Also pins both jobs to non-preempt — preemption forces a full ICA poll restart and wastes hours of DRAGEN runtime. Matches `dragen-unified-dev` and the existing `UploadDataToIca` pattern.

cpg_flow's rerun-skip still consumes `expected_outputs()` at the dispatcher and is unaffected.

## Changes

- `src/dragen_align_pa/utils.py` — generic `write_manifest` / `load_manifest` + `ManifestEntry` type (`dict[str, Path] | Path | None`).
- `src/dragen_align_pa/stages.py` — both stage dispatchers write the manifest then call workers with only `manifest_path`; both gain `job.spot(is_spot=False)`.
- `src/dragen_align_pa/jobs/manage_dragen_pipeline.py`, `manage_dragen_mlr.py` — `run()` takes `manifest_path`; unpacks via `cast` at the boundary.
- `tests/test_manage_dragen_manifest.py` — pins the JSON schema for both stage shapes.
- `tests/conftest.py` — minimal `cpg_utils.config` stubs so the new tests can import `dragen_align_pa.constants`.
- `pyproject.toml` — `testpaths` typo (`test` → `tests`) so the new tests are discovered.

## Test plan

- [x] `pytest tests/` — 3 passed
- [x] `pytest --collect-only` — collects cleanly
- [x] `ruff check` + `ruff format --check` on changed files
- [x] `mypy` on changed files
- [ ] Dry-run a small cohort to confirm manifest is written and the worker reads it back

### DIFF
--- a/.github/workflows/get_version.py
+++ b/.github/workflows/get_version.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import os
 import re
 import subprocess
@@ -34,8 +35,6 @@ def get_next_version_tag(folder: str, version: str) -> str:
     full_image_name_archive = f'{base_image_path_archive}/{folder}'
 
     tags_list = []
-    import logging
-
     logging.basicConfig(level=logging.ERROR)
     for full_image_name in [full_image_name_prod, full_image_name_archive]:
         cmd = [
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,7 @@ repos:
     # Ruff version.
     rev: v0.15.0
     hooks:
-      - id: ruff
+      - id: ruff-check
 
   # Static type analysis (as much as it's possible in python using type hints)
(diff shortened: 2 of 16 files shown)


michael-harper (approved):
LGTM

PR #59, opened 2026-05-26
Make the DRAGEN CYP2D6 caller toggleable via config

Alexander-Stuckey:
## Summary

The CYP2D6 caller was hardcoded on inside `_build_common_parameters` (`enable_cyp2d6 = 'true'`). This was because we thought that we would always run it. 
However, some samples could have a hom-deletion in CPY2D6, or otherwise poor coverage, and this would cause the pipeline to _always_ fail.
Now expose it via config, and default to `True` (enabled).

## Changes

- `src/dragen_align_pa/jobs/run_align_genotype_with_dragen.py` — `_build_common_parameters` reads `enable_cyp2d6` from config (default `True`) and maps to the `'true'`/`'false'` string ICA expects.
- `config/dragen_align_pa_defaults.toml` — adds `enable_cyp2d6 = true` under `[workflow]` with a short comment.
- `tests/test_enable_cyp2d6.py` — covers default-on, explicit-on, and disabled-via-config.

## Test plan

- [x] `pytest tests/` — 6 passed (3 new + 3 existing)
- [x] `ruff check` + `ruff format --check`
- [x] `mypy`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM australia-southeast1-docker.pkg.dev/cpg-common/images/cpg_hail_gcloud:0.2.138.cpg2-1
 
 # Dragen align pa pipeline version.
-ENV VERSION=3.4.0
+ENV VERSION=3.4.1
 
 ARG ICA_CLI_VERSION="2.45.0"
 ARG SOMALIER_VERSION="0.3.1"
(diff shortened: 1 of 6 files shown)


joshuamschmidt (approved):
LGTM

PR #60, opened 2026-05-26
Fix AR GUID preservation bug between the pipeline-id file and ICA userReference

Alexander-Stuckey:
## Summary

`force_resubmit` was preserving the stale AR GUID from the existing pipeline-id file, while `submit_dragen_run` independently called `try_get_ar_guid()` to build the ICA `userReference`. The two diverged, so the JSON stored one GUID and the actual ICA folder was named with another — every downstream stage (`download_specific_files_from_ica`, `download_md5_results`, `manage_dragen_mlr`) then 404s on a path that doesn't exist. This stamps the same AR GUID on both sides by construction. Also lifts the previously-commented-out CI workflow from `dragen-unified-dev` so pytest actually runs on PRs.

## Changes

- `src/dragen_align_pa/jobs/ica_pipeline_manager.py` — drop the read-and-preserve of the old `ar_guid` in the `force_resubmit` branch (just delete the stale files); change `submit_function_factory` contract to `(sg_name, ar_guid) -> Callable`, pick the AR GUID once per iteration and reuse it for both first-submit and retry so the dead `target.ar_guid or initial_ar_guid` fallback is gone.
- `src/dragen_align_pa/jobs/run_align_genotype_with_dragen.py` — `submit_dragen_run` and `run` take an `ar_guid` parameter and build `userReference` from it instead of calling `try_get_ar_guid()`; drop the now-unused import.
- `src/dragen_align_pa/jobs/manage_dragen_pipeline.py` — thread `ar_guid` through `_submit_new_ica_pipeline` into `submit_dragen_run`.
- `src/dragen_align_pa/jobs/manage_dragen_mlr.py` — accept the new factory arg and explicitly discard it (MLR reads its `ar_guid` from the upstream Dragen pipeline-id file, not from the env).
- `src/dragen_align_pa/jobs/manage_md5_pipeline.py` — accept the new factory arg and discard it (MD5 binds its own `ar_guid` eagerly above the factory).
- `tests/test_ica_pipeline_manager_ar_guid.py` — pin all three contracts: `force_resubmit` must not preserve, the factory must receive `(sg_name, ar_guid)`, and `submit_dragen_run`'s `userReference` must come from its argument.
- `.github/workflows/test.yaml` — lifted from `dragen-unified-dev` so the test suite actually runs on PRs (the existing file was fully commented out).

## Test plan

- [x] `pytest` — 6 passed (3 new + 3 existing)
- [x] `ruff check` at pinned `v0.15.0` — clean on all touched files
- [x] `mypy` at pinned `v1.19.1` — clean on all touched files
- [x] CI passes on this PR

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,29 +1,35 @@
-# name: Run Pytest
-# on:
-#   # Running on pull-requests and new pushes to main
-#   pull_request:
-#   push:
-#     branches:
-#       - main
+name: Run Pytest
+on:
+  # Running on pull-requests and new pushes to main
+  pull_request:
+  push:
+    branches:
+      - main
 
-# jobs:
-#   test:
-#     runs-on: ubuntu-latest
-#     defaults:
-#       run:
-#         shell: bash -l {0}
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash -l {0}
 
-#     steps:
-#     - uses: actions/checkout@v4
+    steps:
+    - uses: actions/checkout@v5
 
-#     - uses: actions/setup-python@v4
-#       with:
-#         python-version: '3.11'
-#         cache: 'pip'
-#         cache-dependency-path: pyproject.toml
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.11'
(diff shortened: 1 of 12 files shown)


michael-harper (approved):
LGTM

PR #61, opened 2026-05-28
Retry check_ica_pipeline_status on transient ICA 429/503

Alexander-Stuckey:
## Summary

When ICA's API returns HTTP 429 ("too many requests") or 503, the cohort monitor jobs currently crash on the first occurrence. This PR makes the status-polling call retry transient errors instead of dying.

## What it does

- Retries the per-pipeline status check (\`ica_api_utils.check_ica_pipeline_status\`) on HTTP 429 or 503, up to 5 attempts.
- Backs off between retries with jitter: \`wait_exponential(min=2, max=30) + wait_random(0, 5)\` seconds.
- Leaves all other HTTP errors (404 not-found, 500 server error, etc.) unchanged — they still raise on the first occurrence.
- Stops re-wrapping \`icasdk.ApiException\` at three call sites in \`ica_api_utils.py\` (more on this below).

## Why now

- Production logs show \`ManageDragenPipeline\` and \`ManageDragenMlr\` being torn down by single transient 429s during cohort polling. The fix is small and lands behaviour we already want.
- Required for the upcoming parallel-polling redesign (next PR): once 16 workers run concurrently, a 429 hitting all of them in lockstep would be its own self-inflicted second wave. The \`wait_random\` jitter spreads retries out.

## Why the re-wrap had to go first

The three wrappers in \`ica_api_utils.py\` re-raised exceptions as \`raise icasdk.ApiException(f'... {e}') from e\`. \`ApiException.__init__\`'s first positional argument is \`status=\`, so that f-string was being assigned to \`.status\` — making it a string instead of an HTTP code. Any predicate of the shape \`isinstance(e, ApiException) and e.status in (429, 503)\` would never match. Dropping the re-wrap (log the context, then bare \`raise\`) preserves the original exception so the retry predicate works.

## Changes

- \`src/dragen_align_pa/ica_api_utils.py\`:
  - Drop the f-string re-wrap in \`check_ica_pipeline_status\`, \`check_object_already_exists\`, and \`submit_nextflow_analysis\` — log context, then bare \`raise\`.
  - Add \`_is_retryable_ica_error\` predicate (returns True for \`ApiException\` with \`.status\` 429 or 503).
  - Decorate \`check_ica_pipeline_status\` with \`@retry(...)\`.
  - Add a single \`# pyright: ignore[reportAttributeAccessIssue]\` to silence Pyright on the predicate's \`.status\` access (Pyright can't resolve \`icasdk\` from the project venv; same suppression style already used elsewhere in this file).
- \`tests/test_ica_api_utils.py\`: regression tests covering
  - \`.status\` survives the wrapper (originally was being clobbered)
  - retry-then-succeed on 429
  - retry-then-succeed on 503
  - no-retry on 404 (one call, then propagates)
  - give-up after 5 persistent 429s (5 calls, then propagates the original)

## Test plan

- [x] \`pytest tests/test_ica_api_utils.py\` — 13 passed
- [x] \`uvx ruff@0.15.0 check src/dragen_align_pa/ica_api_utils.py tests/test_ica_api_utils.py\` — exit 0
- [x] \`uvx mypy@1.19.1 src/dragen_align_pa/ica_api_utils.py\` — exit 0
- [ ] Confirm on a real cohort run that a single 429 no longer tears down the monitor jobs.

Same retry decorator should later wrap \`submit_nextflow_analysis\` and the other ICA REST calls in the same file; tracked as a follow-up.

### DIFF
--- a/src/dragen_align_pa/ica_api_utils.py
+++ b/src/dragen_align_pa/ica_api_utils.py
@@ -128,7 +128,7 @@ def _is_retryable_ica_error(exc: BaseException) -> bool:
     mode); 503 = ICA backend unavailable. 404/500/etc propagate immediately
     — retrying a permanent error just delays the real signal.
     """
-    return isinstance(exc, ApiException) and exc.status in (429, 503)
+    return isinstance(exc, ApiException) and exc.status in (429, 503)  # pyright: ignore[reportAttributeAccessIssue]
 
 
 @retry(
--- a/src/dragen_align_pa/ica_api_utils.py
+++ b/src/dragen_align_pa/ica_api_utils.py
@@ -20,9 +20,11 @@ from icasdk.model.create_nextflow_analysis import CreateNextflowAnalysis
 from loguru import logger
 from tenacity import (
     retry,
+    retry_if_exception,
     retry_if_exception_type,
     stop_after_attempt,
     wait_exponential,
+    wait_random,
 )
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


src/dragen_align_pa/ica_api_utils.py
```diff
@@ -117,6 +119,29 @@ def get_ica_api_client() -> Iterator[ApiClient]:
...
+
+@retry(
+    retry=retry_if_exception(_is_retryable_ica_error),
+    stop=stop_after_attempt(5),
+    # wait_random(0, 5) jitter is non-negotiable at 16-wide fan-out: without
+    # it, every worker that sees 429 backs off in lockstep and re-hits ICA
+    # together (a self-inflicted second wave). max=30 (not 60) keeps a
+    # single id's worst-case retry budget (~60s) comfortably inside MLR's
+    # 330s outer cycle.
+    wait=wait_exponential(multiplier=1, min=2, max=30) + wait_random(0, 5),
```
joshuamschmidt: maybe consider `wait_random_exponential` as a defense against upstream changes to increased parallelism, as the range `[min, max]` increases on each retry.
```
wait_random_exponential(multiplier=1, max=30) + wait_random(2, 3)
```
the `+ wait_random(2, 3)` stops small potential instant hammering, as wait_random_exponential lower bound is `[0`

Alexander-Stuckey: Makes sense to me

joshuamschmidt (approved):
LGTM. 
I made one suggestion re jittering strategy which is not a blocker

src/dragen_align_pa/ica_api_utils.py
joshuamschmidt: I think I edited this comment after you replied sorry.

PR #62, opened 2026-05-31
Delete legacy per-SG DRAGEN submitter

michael-harper:
## Summary

- Deletes `src/dragen_align_pa/jobs/run_align_genotype_with_dragen.py` (316 lines). The per-SG submitter was superseded by `submit_dragen_batch.py` plus the stage rewiring in PR #52; nothing imports it.
- Drops the orphaned `[ica.pipelines] cram` and `[ica.pipelines] fastq` keys from `dragen_align_pa_defaults.toml` (only the legacy submitter read them; the unified pipeline uses `[dragen_align_pa.manage_dragen_pipeline] pipeline_id`).
- Rewrites the docstring at the top of `submit_dragen_batch.py` to drop the now-dangling reference.

## Wire WGS QC coverage regions into the unified pipeline

Supersedes the earlier "keep `coverage_region_1`/`_2` with a TODO" plan — the WGS QC coverage BEDs are now actually wired in:

- Registers all three QC files in `dragen_align_pa.constants.ICA_FILE_IDS` under their real ICA basenames (`SNP_NCBI_GRCh38.vcf`, `wgs_coverage_regions.hg38_minus_N.interval_list.bed`, `acmg59_allofus_19dec2019.GRC38.wGenes.NEW.bed`), so they resolve by name like the existing target BEDs.
- Replaces the single global `[ica.qc] coverage_region_beds` list with per-seqtype blocks `[ica.qc.genome]` / `[ica.qc.exome]`. The submitter now reads the block matching `workflow.sequencing_type`, so WGS regions only go to WGS runs and exome regions only to exome runs (previously the one global list was applied to every run regardless of seqtype).
- Converts `[ica.qc] cross_cont_vcf` from a raw `fil.…` ID to a basename resolved via `ICA_FILE_IDS`, consistent with the BEDs. A typo/unstaged file now fails fast at submitter startup instead of as an opaque mid-run ICA error.
- Removes the now-dead `coverage_region_1` / `coverage_region_2` raw-ID keys (nothing read them).

### QC file IDs were refreshed

The `fil.…` IDs for all three QC files **changed** from what the TOML previously held. These are the *same* files, they were re-uploaded to ICA at some point (during debugging of an unrelated issue), which minted new IDs. The old IDs in the prior config were stale. This PR uses the current authoritative IDs pulled from ICA:

| File | Old (stale) ID | New ID |
|---|---|---|
| `SNP_NCBI_GRCh38.vcf` (cross_cont_vcf) | `fil.1a7a2d04…b5b0` | `fil.fd99781d…afe1ac` |
| `wgs_coverage_regions.hg38_minus_N.interval_list.bed` | `fil.ad60897b…ba0a20` | `fil.434cd66e…159355` |
| `acmg59_allofus_19dec2019.GRC38.wGenes.NEW.bed` | `fil.c4ade57f…ba0a20` | `fil.d37b27f6…298bbd` |

## Test plan

- [ ] CI runs pre-commit + pytest cleanly.
- [ ] Spot-check that no other config readers reach for `[ica.pipelines] cram` / `fastq` (already verified locally with grep — only `run_align_genotype_with_dragen.py` did).
- [ ] Confirm `submit_dragen_batch` still imports cleanly after the docstring change.
- [ ] New tests cover per-seqtype `coverage_region_beds` resolution and name-based `cross_cont_vcf` (resolve + fail-fast on typo).


Alexander-Stuckey (approved):
This all looks good to me

PR #63, opened 2026-06-01
Bump reheader storage

Alexander-Stuckey:
# Purpose

  - Very occasionally, there is a gvcf that is very large, so reheadering it takes more space than the default allocation. Adding a config toggle to increase storage in these cases.


michael-harper (approved):
Will we know ahead of time how large they are or do we have to wait for failures?

PR #65, opened 2026-06-02
Retry check_ica_pipeline_status on transient ICA 429/503

Alexander-Stuckey:
## Summary

I need to add the same backoff and retry settings from dev into main, as downloading the non CRAM / gvcf files from ICA is causing more ICA API timeouts.

## Changes

- `src/dragen_align_pa/ica_api_utils.py`: add `_is_retryable_ica_error` predicate; build a call-time `Retrying` (`_ica_status_retrying`) driven by `[ica.retry] max_retries` (default 10) with `wait_random_exponential(min=2, max=30) + wait_fixed(2)`; stop the `.status`-clobbering re-wrap in all three wrappers (`check_ica_pipeline_status`, `check_object_already_exists`, `submit_nextflow_analysis`) — log context, re-raise the original exception.
- `config/dragen_align_pa_defaults.toml`: document the new `[ica.retry] max_retries` knob.
- `pyproject.toml`: add `tenacity` as an explicit dependency (was only transitive); add a `tests/**` ruff per-file-ignore set (PLR2004/ANN202/…) so test code lints clean.
- `tests/test_ica_api_utils.py` (new): `.status` survives the wrapper; 429/503 retry-then-succeed; 404 no-retry; persistent-429 gives up after 11 attempts (10 retries); retry count is configurable. An autouse fixture stubs tenacity's sleep so retry tests don't wait.

## Test plan

- [x] `pytest -q` (16 passed)
- [x] `ruff@0.15.0 check src/dragen_align_pa tests`
- [x] `mypy@1.19.1 src/dragen_align_pa` (with `types-requests`, as the pre-commit hook installs)
- [ ] Confirm on a real run that an injected/observed ICA 429 is absorbed without tearing down the monitor, and that `[ica.retry] max_retries` in the run TOML changes the attempt count.

## Notes

`max_retries` is retries *after* the initial attempt (total tries = `max_retries + 1`). Worst-case backoff is ~32s per retry at the cap, so a large value can exceed a tight polling cycle — tune with that in mind.

### DIFF
--- a/config/dragen_align_pa_defaults.toml
+++ b/config/dragen_align_pa_defaults.toml
@@ -63,6 +63,14 @@ reference_tags = ['test_reference_tags']
 [ica.api]
 batch_size = 30
 
+# Transient-error retry for ICA status polling (check_ica_pipeline_status).
+# max_retries = retries AFTER the initial attempt, so total tries = max_retries + 1.
+# Read at call time, so this can be tuned per-run without rebuilding the image.
+# NB: worst-case backoff is ~32s per retry at the cap, so a large value can
+# exceed a tight polling cycle.
+[ica.retry]
+max_retries = 10
+
 # The pipeline ID is from ICA and is specific to the exact pipeline being run. This should not need to be changed.
 [ica.pipelines]
 dragen_version = 'dragen_3_7_8'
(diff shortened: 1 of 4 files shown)


joshuamschmidt (approved):
more or less the same as #61 but with extra plumbing? LGTM

michael-harper (approved):
Is `Tenacity` used elsewhere in CPG? I'm not familiar with it.

PR #66, opened 2026-06-02
Apply ICA 429/503 retry to all data-plane calls

Alexander-Stuckey:
## Summary

The tenacity retry added in 762f842 was invoked at a single call site (`check_ica_pipeline_status` → `get_analysis`), and I forgot to wire it in for all the other API calls, especially the ones where it matters most (like downloading bulk data). This meant that all the download calls had no retires.

## Changes

- `src/dragen_align_pa/ica_api_utils.py`: rename `_ica_status_retrying` → `_ica_retrying` (the backoff is generic, not status-specific); add a public `ica_retry(fn, *args, **kwargs)` helper that wraps one SDK call in the controller; add a `before_sleep` hook so each retry is logged; route `get_analysis`, `get_project_data_list` (×3), and `create_nextflow_analysis` through it.

- `src/dragen_align_pa/ica_utils.py`: route `create_data_in_project`, both `create_download_url_for_data` calls (incl. the one in the production traceback), and `get_project_data_list` through `ica_retry`.

- `src/dragen_align_pa/jobs/download_md5_results.py`: route `get_project_data_list` and `create_download_url_for_data` through `ica_retry`.

- `tests/test_ica_api_utils.py`: add coverage for the `ica_retry` helper (retries transient, not 404) and for retry on `submit_nextflow_analysis` and `check_object_already_exists`.

- `tests/test_ica_utils.py`: new file pinning the originally-failing path — `stream_ica_file_to_gcs` retries a 429 on the download-URL call and gives up after persistent 429s.

## Test plan

- [x] `pytest tests/` (22 passed)
- [x] `ruff check` (v0.15.0, pinned) — clean
- [x] `pre-commit run` (mypy v1.19.1, ruff, cpg-id-checker) — pass

### DIFF
--- a/src/dragen_align_pa/ica_api_utils.py
+++ b/src/dragen_align_pa/ica_api_utils.py
@@ -7,8 +7,8 @@ wrappers around specific API endpoints.
 import contextlib
 import functools
 import json
-from collections.abc import Iterator
-from typing import TYPE_CHECKING, Any, Final, Literal
+from collections.abc import Callable, Iterator
+from typing import TYPE_CHECKING, Any, Final, Literal, TypeVar
 
 import icasdk
 from cpg_utils.config import config_retrieve
@@ -19,6 +19,7 @@ from icasdk.exceptions import ApiException
 from icasdk.model.create_nextflow_analysis import CreateNextflowAnalysis
 from loguru import logger
 from tenacity import (
+    RetryCallState,
     Retrying,
     retry_if_exception,
     stop_after_attempt,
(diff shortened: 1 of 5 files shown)


michael-harper (approved):
Looks good to me!
One comment regarding `stream_ica_file_to_gcs` and is out of scope of this PR; we use `blob.open('wb')` which implements a resumable upload but reading from ICA's pre-signed URL, is not covered by any retry, a drop there raises `requests.RequestException` straight out of the loop. `ica_retry` can't help either: it only matches `icasdk.ApiException` `429/503`, and the streaming call raises requests errors against the object store, not the ICA API.

This may be difficult to find/unpack so definitely for future work, but does ICA provide any resumable avenues here?

PR #67, opened 2026-06-03
Generalise ICA 429/503 retry to all data-plane calls

Alexander-Stuckey:
## Summary

Ports the retry generalisation shipped to `main` in #66 onto `dev`. On `dev` only `check_ica_pipeline_status` was retried — via a `@retry` decorator hardcoded to 5 attempts — so every other ICA call (download-URL minting, data listing, pipeline submission) had no retry, and a transient 429 propagated unhandled and tore down the job. This routes all data-plane SDK calls through one shared, config-tunable retry.

## Changes

- `src/dragen_align_pa/ica_api_utils.py`: replace the `@retry` decorator on `check_ica_pipeline_status` with a shared controller — `_ica_retrying()` builds the tenacity `Retrying` (`[ica.retry] max_retries`, default 10; jittered exponential backoff + 2s floor; `before_sleep` logging) and a public `ica_retry(fn, *args, **kwargs)` helper wraps a single SDK call. Route `get_analysis`, `get_project_data_list` (×3), and `create_nextflow_analysis` through it. The separate `get_ica_secrets` Secret Manager retry is left untouched.
- `src/dragen_align_pa/ica_utils.py`: route `create_data_in_project`, both `create_download_url_for_data` calls, and `get_project_data_list` through `ica_retry`.
- `src/dragen_align_pa/jobs/download_md5_results.py`, `jobs/parse_passfail.py`, `jobs/manage_md5_pipeline.py`: route their `get_project_data_list` / `create_download_url_for_data` calls through `ica_retry`.
- `tests/test_ica_api_utils.py`: update the give-up test to the new default (10 retries → 11 attempts); add the configurable-count test, `ica_retry` helper tests, and retry coverage for `submit_nextflow_analysis` and `check_object_already_exists`.
- `tests/test_ica_utils.py`: new file pinning `stream_ica_file_to_gcs` — retries a 429 on the download-URL call and gives up after persistent 429s.

## Test plan

- [x] `pytest tests/` (174 passed)
- [x] `ruff check` (v0.15.0, pinned) — clean
- [x] `pre-commit run` (mypy v1.19.1, ruff, cpg-id-checker) — pass

PR #68, opened 2026-06-03
Batch-mint ICA download URLs to cut per-file 429s

Alexander-Stuckey:
## Summary

The bulk-output download minted one pre-signed URL per file (`:createDownloadUrl`), so a folder with N outputs made N rate-limited calls — the dominant 429 source at ~500-SG concurrency. This switches to the batch `:createDownloadUrls` endpoint, which returns every file's URL in one call. (A pre-signed URL is per-object, so this is N URLs in one response, not a single folder URL — the fewest API calls available without abandoning the stream-to-GCS-while-hashing design.)

Stacked on #67 (`ica_retry`); base will be retargeted to `dragen-unified-dev` once #67 merges.

## Changes

- `src/dragen_align_pa/ica_utils.py`: add `batch_create_download_urls` — one `ica_retry`-wrapped `create_download_urls_for_data` call over a list of `dataIds`, returning a `{dataId: url}` map (empty input short-circuits with no API call). Add an optional `download_url` param to `stream_ica_file_to_gcs`: when supplied the per-file mint is skipped; otherwise it mints as before (backward-compatible for `download_specific_files_from_ica`).
- `src/dragen_align_pa/jobs/download_ica_pipeline_outputs.py`: chunk the file list and batch-mint per chunk just-in-time (URLs fresh when streamed); a file id absent from the batch response falls back to a per-file mint. Chunk size via `[ica.download] url_batch_size`.
- `config/dragen_align_pa_defaults.toml`: add `[ica.download] url_batch_size = 50` with a note on the URL-TTL trade-off.
- `tests/test_ica_utils.py`: cover the `{dataId: url}` map, empty-input short-circuit, 429 retry on the batch mint, and that `stream_ica_file_to_gcs` skips minting when a URL is supplied.

Net: per-file ICA API calls for the job drop from `1 list + N mints` to `1 list + ceil(N / url_batch_size) batch mints`.

## Test plan

- [x] `pytest tests/` (178 passed)
- [x] `ruff check` (v0.15.0, pinned) — clean
- [x] `pre-commit run` (mypy v1.19.1, ruff, cpg-id-checker) — pass
- [x] Confirm ICA pre-signed URL TTL comfortably exceeds the time to stream one `url_batch_size` chunk; tune the default down if not.


joshuamschmidt (approved):
LGTM

PR #72, opened 2026-06-16
Panel of normals Generation Script

michael-harper:
## Context

A Panel Of Normals (PON) needs to be generated for the exome CNV calling pipeline (genome CNV calling uses `self-normalisation` and doesn't need a PON).

DRAGEN handles the generation of the PON internally (per-target median across the listed samples, preconditioning, normalisation etc.), all that needs to be supplied is a single text file containing gc-corrected target counts from previous DRAGEN runs: `gs://cpg-bioheart-test/ica/dragen_3_7_8/output/dragen_metrics/<sgid>/<sgid>.target.counts.gc-corrected.gz`. 

The DRAGEN guide requires the panel samples to share identical library prep + sequencing workflow + interval settings as the case (p.129-130). 

The panel is built once per stratum and then reused for every future case sample that matches. At run time we route each case to the panel matching its own kit e.g. Twist case → the twist panel; a CREv2-nswhp-late case → the crev2-nswhp-late panel (or, if we want to break down by sex as well: e.g. a male Twist case → the twist-male panel; a female CREv2-nswhp-late case → the crev2-nswhp-late-female panel).

So one PON per library-prep/sequencing stratum, ×2 for sex if sex-chromosome CNVs are needed.

## This PR
Per panel cohort: copies existing gc-corrected target count files to user-specified directory on GCS for provenance, copies them to `references` directory on ICA, writes ICA paths into txt file.

## Wiring into Pipeline
1. Run this PR's script → it uploads the ~N count files + normals.txt to ICA and prints a paste-ready `ICA_FILE_IDS` snippet.
2. Create PR to load those `fil.…` IDs into `ICA_FILE_IDS` (`constants.py`), same registry the BEDs already use.
3. In the run config:
  - add the count-file basenames and the `normals.txt` basename to `[presets.<seqtype>].additional_files`;
  - add `--cnv-normals-list <panel>.normals.txt` (literal basename) to that preset's additional_args.
  - Exome already has `--cnv-enable-self-normalization false`, so it's done. For the WGS wiring test, also flip the genome preset from `true` → `false` (in a per-run override, not committed defaults, it would affect all WGS runs).


joshuamschmidt (approved):
LGTM

scripts/build_cnv_panel_of_normals.py
```diff
@@ -0,0 +1,222 @@
...
+from typing import Literal
+
+from icasdk.apis.tags import project_data_api
+from loguru import logger
+
+from dragen_align_pa import ica_api_utils, ica_cli_utils, utils
+
+# DRAGEN names gc-corrected counts <SG>.target.counts.gc-corrected.gz. The PON
+# list points at these (guide p.130 example, and gc-corrected is recommended
+# for WGS downstream). Override via --counts-suffix for plain counts on WES
```
Alexander-Stuckey: This probably holds for WES as well, but worth checking.
For WGS we don't even have to do this procedure, so we don't really care.

scripts/build_cnv_panel_of_normals.py
```diff
@@ -0,0 +1,222 @@
...
+        return existing['id']
+
+    local_path = os.path.join(ica_local_dir, basename)
+    utils.run_subprocess_with_log(
+        ['gcloud', 'storage', 'cp', gcs_counts_path, local_path],
+        f'Download {basename} from GCS',
+    )
+    ica_cli_utils.upload_local_file(local_path, ica_folder_path)
+
+    return _require_file_id(api_instance, path_params, ica_folder_path, basename)
```
Alexander-Stuckey: I've made a function n `ica_utils` -> `ica_utils.wait_for_file_available`, as trying to return immediately causes a race condition where ICA might not have registered that the file has been uploaded in the first few seconds. I would recommend using it, as this race condition has bitten me before.

scripts/build_cnv_panel_of_normals.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+def build_panel(
+    panel_name: str,
+    sequencing_groups: list[str],
+    ica_reference_folder: str,
+    provenance_prefix: str | None,
+    counts_suffix: str,
+) -> dict[str, str]:
+    """Run steps 1-5 for one panel. Returns {basename: ica_file_id} for every
```
Alexander-Stuckey: Worth updating the docstrings to match the proposed style guide: https://github.com/populationgenomics/popgen-collaborative-ai-coding-template/blob/onboarding-resource/docs/style/python.md#docstrings

scripts/build_cnv_panel_of_normals.py
```diff
@@ -0,0 +1,222 @@
...
+            )
+
+        list_basename = f'{panel_name}.normals.txt'
+        list_local = os.path.join(ica_local_dir, list_basename)
+        with open(list_local, 'w') as fh:
+            for sg in sequencing_groups:
+                fh.write(_counts_basename(sg, counts_suffix) + '\n')
+        logger.info(f'[{panel_name}] wrote {list_basename} with {len(sequencing_groups)} entries.')
+        ica_cli_utils.upload_local_file(list_local, ica_folder_path)
+        file_ids[list_basename] = _require_file_id(api_instance, path_params, ica_folder_path, list_basename)
```
Alexander-Stuckey, line 316: Race condition here again

scripts/build_cnv_panel_of_normals.py
```diff
@@ -0,0 +1,222 @@
...
+def _print_registration_snippet(panel_name: str, file_ids: dict[str, str]) -> None:
+    """Emit a paste-ready ICA_FILE_IDS block for constants.py."""
+    logger.info(
+        f'Panel "{panel_name}" built. Add these to dragen_align_pa.constants.ICA_FILE_IDS '
+        f'so the submitter can resolve them by name:',
+    )
+    print(f'\n    # --- CNV PON: {panel_name} ---')
+    for basename, fid in file_ids.items():
+        print(f"    '{basename}': '{fid}',")
+    print()
```
Alexander-Stuckey: For each panel (4x), this is going to be 101 entries (50 male, 50 female + 1 list file). 
Not a blocker, but needs to be formatted nicely. I would recommend perhaps
`json.dump`, so we get a nicely formatted `JSON` block to paste into `constants.py`

Alexander-Stuckey (changes requested):
Some thoughts here, mostly good otherwise

scripts/build_cnv_panel_of_normals.py
michael-harper: Very handy

scripts/build_cnv_panel_of_normals.py
michael-harper: Dumping out json now

scripts/build_cnv_panel_of_normals.py
michael-harper: Yes slightly confusing now that I'm looking back at it

michael-harper: @joshuamschmidt can you confirm that our targets have >200k?

joshuamschmidt: @michael-harper tbh I am a little confused here as to the exact context you and alex are discussing. 
But from first principles, I'd say that GC correction is pretty standard for WES CNV and I would have expected we used it, and i don't have intuition as to why the n of targets bears on that decision.

> can you confirm that our targets have >200k?  

twist-vcgs: 229,273
crevb2: 275,017

michael-harper: The full paragraph from DRAGEN's user guide:
The GC bias correction module immediately follows the target counts stage and operates on the
.target.count file. GC bias correction generates a GC bias corrected version of the file, which has a
.target.counts.gc-corrected.gz extension in the file name. The GC bias corrected versions are
recommended for any downstream processing when working with WGS data. For WES, if there are
enough target regions, then the GC bias corrected counts can also be used.
Typical capture kits have over 200,000 targets spanning the regions of interest. If your BED file has
fewer than 200,000 targets, or if the target regions are localized to a specific region in the genome
(such that GC bias statistics may be skewed), then GC bias correction should be disabled.

joshuamschmidt: ah that makes sense now -> need enough regions to get a good model of GC bias. 200K is pretty conservative, but i guess this is to stop user error for people using GC correction on targeted gene panels.

Alexander-Stuckey (approved):
Looks good to me

PR #73, opened 2026-06-24
Allow dragen cram input

Alexander-Stuckey:
# Purpose

  - Dragen CRAMs are stored at `<bucket>/ica/...`, which is NOT the path that sequencing_group.cram` resolves to. If we need to rerun Dragen using a Dragen generated CRAM file, this change enables that.

## Proposed Changes

  - Config setting to switch between CRAMs that are produced by Dragen and live under `<bucket>/ica/...` and others that live under the `sequencing_group.cram` path.
  - Update the docstring for the `_setup_paths` function to comply to the new styleguide.
  - Fix the `wait_for_file_available` function to work correctly on `FileNotFound`, not always.

## Checklist

- [ ] Related GitHub Issue created
- [ ] Tests covering new change
- [ ] Linting checks pass

### DIFF
--- a/src/dragen_align_pa/jobs/upload_data_to_ica.py
+++ b/src/dragen_align_pa/jobs/upload_data_to_ica.py
@@ -27,8 +27,8 @@ def _setup_paths(
 ) -> dict[str, str]:
     """Resolve the GCS source, local scratch, and ICA destination paths for the upload.
 
-    In DRAGEN mode (``use_dragen_crams``) the source is the DRAGEN realignment
-    output CRAM; otherwise it is the sequencing group's own CRAM.
+    Allows for using Dragen generated CRAMs if the config setting
+    'dragen_align_pa.upload_data_to_ica.use_dragen_crams' is true.
 
     Args:
         sequencing_group: The sequencing group whose CRAM is being uploaded;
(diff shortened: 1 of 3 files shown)


src/dragen_align_pa/jobs/upload_data_to_ica.py
```diff
@@ -11,35 +11,61 @@
...
-    # Ensure we get the .cram path, not .crai
-    gcs_base_path = str(sequencing_group.cram)
-    if gcs_base_path.endswith('.cram.crai'):
-        gcs_cram_path = gcs_base_path.removesuffix('.crai')
-    elif not gcs_base_path.endswith('.cram'):
-        raise ValueError(
-            f'Unexpected path for sequencing_group.cram: {gcs_base_path}',
+    if cpg_utils.config.config_retrieve(['dragen_align_pa', 'upload_data_to_ica', 'use_dragen_crams'], False):
+        gcs_cram_path: cpg_utils.Path | str = (
+            sequencing_group.dataset.prefix()
```
joshuamschmidt, line 51: not a blocker, but I guess there is the minor potential for the path here, and the path constructed for CRAM download could drift?

joshuamschmidt (approved):
i think this is good.

src/dragen_align_pa/jobs/upload_data_to_ica.py
Alexander-Stuckey: Yeah there is. There's an entirely separate bit of work that needs to be done to canonically create paths and then share them around, but that's bigger than this change

PR #74, opened 2026-06-25
Fix missing cov reports bed output

Alexander-Stuckey:
# Purpose

  - The new unified pipeline changed how QC coverage reports were generated (again). This change adds back in the missing report.
  
## Proposed Changes

  - Add back in flag to generate missing QC report


michael-harper (approved):
LGTM

PR #76, opened 2026-06-28
Cast lanes to str, as the format in the manifest changed

Alexander-Stuckey:
# Purpose

  - The manifest file parser changed the data in the lane column from Lane01 -> 1, thus breaking the pipeline. This fix casts the lane column to a `str` when reading in the manifest.

## Proposed Changes

  - Cast lane to `str`, along with other manifest columns that could break the pipeline.


joshuamschmidt (approved):
LGTM

PR #77, opened 2026-06-29
Normalise ICA path construction

Alexander-Stuckey:
Normalise the ICA path construction to ensure leading and trailing forward slash in the parentPath

# Purpose

  - ICA paths in API requests strictly enforce starting and ending with `/` in the `parentPath` attribute. This was inconsistently applied over the codebase, and lead to issues (although not in test?)

## Proposed Changes

  - Normalise the `parentPath` in the helper functions, so that even if the code forgets to construct the path correctly, it is ok.
  -

## Checklist

- [x] Tests covering new change
- [ ] Linting checks pass

### DIFF
--- a/src/dragen_align_pa/ica_api_utils.py
+++ b/src/dragen_align_pa/ica_api_utils.py
@@ -154,8 +154,7 @@ def _log_ica_retry(retry_state: RetryCallState) -> None:
     fn_name = getattr(retry_state.fn, '__name__', repr(retry_state.fn))
     sleep = retry_state.next_action.sleep if retry_state.next_action else 0.0
     logger.warning(
-        f'ICA {fn_name} returned {status}; retrying '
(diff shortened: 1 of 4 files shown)


src/dragen_align_pa/ica_api_utils.py
joshuamschmidt: minor - but i gather this would need to be the following to exactly match the path queried?

```
'parentFolderPath': '/' + parent_folder_path.strip('/') + '/'
f'File not found in ICA: {parent_folder_path}{file_name}'
```

joshuamschmidt (approved):
LGTM - one minor point that is not blocking

src/dragen_align_pa/ica_api_utils.py
Alexander-Stuckey: Good catch

PR #78, opened 2026-06-29
Suffix panel-of-normals sample identities to avoid caseSampleNotInPoN

michael-harper:
Dragen errors out when file names in panel of normals txt file match those in the current pipeline run. This script renames the files (gc-correct counts files `.tsv`) and their contents by adding a user supplied suffix (defaults to `_pon`) to avoid triggering this clash.

In our test data the contents of the gc-corrected TSV files has a different sequencing group ID than the file name. This is due to the `create_test_subset.py` script that did not reheader the cram when bringing across test data and so the column header has the old sequencing group ID. This script accounts for this difference. 

Relevant Slack thread [here](https://centrepopgen.slack.com/archives/C08JL4YHPNE/p1782700420273949?thread_ts=1782445901.454889&cid=C08JL4YHPNE)


Alexander-Stuckey (approved):
Looks good to me

PR #79, opened 2026-06-29
Add back md5sum calculations for the reheadered MLR gvcfs

Alexander-Stuckey:
# Purpose

  - Reheadering the MLR gvcfs invalidated the ICA md5sums, and they also were not being copied over from `-tmp` (an oversight, but fortunate as they would not have matched). This change re-calculates the md5sum on the reheadered MLR gvcf and persists it to main storage.

## Proposed Changes

  - Add in md5sum calculation after reheadering.

## Checklist

- [x] Linting checks pass
- [x] Checks on `fewgenomes` pass


katiedelange (approved):
LGTM

